PR #320, opened 2026-05-16
fix(artifacts): give create_artifact a CSP-safe React recipe

philmerrell:
## Summary
- React example artifacts rendered blank (HTML shell painted, no content). Root cause: the model's default React pattern loads React/ReactDOM/Babel from `unpkg` and uses `<script type="text/babel">`, all blocked by the artifact iframe CSP, which only permits scripts from `https://esm.sh` and `https://cdn.tailwindcss.com`.
- Updated the `create_artifact` tool docstring with an explicit, CSP-compatible React recipe: React/ReactDOM/`htm` imported as ES modules from `esm.sh`, no Babel, no JSX transpiler. `update_artifact` already defers to these rules.
- Docstring-only change — no infra change, the iframe sandbox/CSP is not widened.

## Notes
- This steers the model strongly but isn't a hard guarantee; the fallback (widening `script-src` to allow `unpkg`/`jsdelivr`) was deliberately not taken to keep the sandbox attack surface small.
- Takes effect once the agents/inference backend is redeployed.

## Test plan
- [ ] Ask the assistant to "create a React demo artifact" and confirm it renders interactive content (not a blank background)
- [ ] Confirm DevTools console shows no CSP `Refused to load the script` violations for the rendered artifact
- [ ] Spot-check that HTML and Markdown artifacts still render unchanged

### DIFF
--- a/backend/src/agents/builtin_tools/artifacts/tools.py
+++ b/backend/src/agents/builtin_tools/artifacts/tools.py
@@ -37,9 +37,31 @@ def make_create_artifact_tool(session_id: str, user_id: str):
         - HTML (default): `content` MUST be a complete standalone HTML
           document (include `<!doctype html>` and a full `<html>` …
           `</html>`). It renders in a sandboxed iframe with a strict CSP:
-          inline `<style>`/`<script>` are allowed, as are scripts from
-          `https://cdn.tailwindcss.com` and `https://esm.sh`. It cannot
-          make network calls.
+          inline `<style>`/`<script>` are allowed, and scripts/modules
+          may ONLY be loaded from `https://cdn.tailwindcss.com` and
+          `https://esm.sh`. It cannot make network calls (`fetch`/XHR
+          are blocked), and scripts from any other CDN (unpkg,
+          jsdelivr, cdnjs, …) are BLOCKED by the CSP — the page will
+          render blank (background only, no content).
+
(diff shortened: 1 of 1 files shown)


philmerrell:
Closing as obsolete — superseded by #326.

This PR's approach was to keep the iframe CSP narrow (esm.sh + tailwind only) and steer the model to an esm.sh + `htm` React recipe, deliberately declining to widen `script-src` to unpkg/jsdelivr "to keep the sandbox attack surface small."

#326 (`fix(artifacts): allow jsdelivr/unpkg CDNs so Chart.js artifacts render`) took the opposite route and merged: `CSP_SCRIPT_SRC` now allowlists `https://cdn.jsdelivr.net` and `https://unpkg.com` (infrastructure/lib/artifacts-stack.ts), and the `create_artifact` docstring was rewritten to document the 4-CDN allowlist. As a result:

- The root cause here (model's default unpkg + Babel React pattern blocked by CSP → blank render) no longer exists; unpkg is now allowlisted.
- This PR's docstring text now claims unpkg/jsdelivr are BLOCKED, which is no longer true, and conflicts with the current docstring.

Note for the team: #320 intentionally avoided widening the sandbox attack surface; #326 widened it. If minimizing the artifact iframe's attack surface is still a goal, that's worth revisiting as its own security-posture discussion rather than via this stale PR.

PR #354, opened 2026-05-19
fix(mcp-apps): align outer CSP + inner mount to ext-apps basic-host reference

philmerrell:
## Summary

- PR #352's blob-URL mount didn't actually escape CSP inheritance — verified on alpha. CSP3 §\"Initialize a Document's CSP list\" treats `about:`, `blob:`, `data:`, `filesystem:` as local schemes that all inherit the embedder's HTTP CSP. The inner iframe still inherited proxy.html's `script-src 'self'`, blocking every bundled-App inline script/style.
- Fix matches `modelcontextprotocol/ext-apps/examples/basic-host` (`src/sandbox.ts` + `serve.ts`): treat CSP inheritance as a feature and ship the outer `proxy.html` with a CSP permissive enough for typical bundled Apps. Per the reference's own comment: \"CSP is enforced via HTTP headers on sandbox.html ... tamper-proof unlike meta tags.\"
- Switch inner-iframe mount from `blob:` URL to `document.write()` (the reference's primary choice — \"Use document.write instead of srcdoc (which the CesiumJS Map won't work with)\"). Falls back to `srcdoc` when the App declared a stricter sandbox without `allow-same-origin`.

## Changes

- **`buildMcpSandboxProxyCsp`**: mirror the reference's default CSP — `'unsafe-inline' 'unsafe-eval' blob: data:` on `script-src`/`style-src`, `worker-src 'self' blob:` for WebGL apps (CesiumJS / Three.js), `connect-src 'self'` (explicitly no `https:` — Apps that need declared `connectDomains` need the dynamic-CSP follow-up tracked separately), `frame-src 'none'`, hardening directives unchanged.
- **`proxy.js`**: switch inner-iframe mount to `document.write()`; default inner sandbox now `allow-scripts allow-same-origin` so `contentDocument` is writable. Safe because the proxy origin is a static CDN with no shared state. Removed `URL.createObjectURL`/`revokeObjectURL` plumbing.
- **`mcp-sandbox-stack.test`**: tighten assertions to lock in the reference-matching default tokens so a future PR can't silently strip them.

## Why this is safe with the broader outer CSP

`proxy.html` itself ships zero inline content (`proxy.js` is external), so `'unsafe-inline'`/`'unsafe-eval'` on the shell can't be exploited unless an attacker can already inject HTML into the static asset — a much larger compromise than this directive's scope. The real containment boundary for untrusted App HTML remains the inner iframe's `sandbox` attribute + cross-origin isolation from the SPA.

## Follow-up

Dynamic per-resource CSP via a CloudFront Function reading `?csp=` (matching the reference's `serve.ts`) is queued as the next PR — required for Apps that declare external `connectDomains`/`resourceDomains` (e.g. Excalidraw's `https://esm.sh`, CesiumJS map-server's OSM tiles). Scoping doc lands with that PR.

## Test plan

- [ ] `cd infrastructure && npm test` passes (asserts new default CSP tokens)
- [ ] Deploy mcp-sandbox stack to dev
- [ ] Invalidate CloudFront `/proxy.html` + `/proxy.js`
- [ ] Open Excalidraw `create_view` in dev — inner App iframe mounts; verify CSP behavior matches expectations (Excalidraw will still log esm.sh blocks until the dynamic-CSP follow-up lands)

### DIFF
--- a/infrastructure/assets/mcp-sandbox/proxy.js
+++ b/infrastructure/assets/mcp-sandbox/proxy.js
@@ -8,27 +8,36 @@
  * This file is served from the dedicated mcp-sandbox origin and runs inside
  * the OUTER iframe the SPA created with sandbox="allow-scripts
  * allow-same-origin". It is the stable cross-origin boundary between the
- * host (ai.client) and the untrusted App View (an inner null-origin srcdoc
- * iframe this script creates).
+ * host (ai.client) and the untrusted App View (an inner iframe this script
+ * creates, mounted via document.write per the ext-apps basic-host
+ * reference). The inner iframe defaults to allow-scripts +
+ * allow-same-origin so document.write can populate it; the App may
+ * override its `_meta.ui.sandbox` to opt back into null-origin (we'll
+ * fall back to srcdoc when contentDocument isn't writable).
  *
  * Responsibilities (spec §"Sandbox proxy"):
  *   3. Announce readiness to the host (ui/notifications/sandbox-proxy-ready).
  *   4. Receive the raw HTML + sandbox/CSP/permissions
  *      (ui/notifications/sandbox-resource-ready).
- *   5. Load the View HTML in the inner iframe with a CSP composed from
- *      _meta.ui.csp plus the spec's deny-by-default fallbacks; map
- *      _meta.ui.permissions onto the inner iframe `allow` attribute.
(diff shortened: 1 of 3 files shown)


philmerrell:
Duplicate of #353, which merged before this PR was opened. Same commit content (`git diff ff072ed9 06bc0500` is empty).

PR #396, opened 2026-05-27
Feature/stack architecture simplification

colinmxs:


philmerrell:
## Code Review — Stack Architecture Simplification

Collapses the 9-stack layout into a single `PlatformStack` of ~30 focused constructs, decouples application code from CFN via bootstrap container/Lambda assets, and consolidates per-stack workflows into `platform.yml` + `backend.yml` + `frontend-deploy.yml`. High-level shape is good; cutover doc is thorough.

### Strengths

- Right architecture for the scaling pain — "CFN updates only when *infrastructure* changes; code rides AWS APIs" matches the actual change frequency. AgentCore Memory's 5-15 min creation no longer rides every backend push.
- Construct decomposition is well done — single responsibility, typed `public readonly` fields, SSM publication, isolated unit tests.
- Bootstrap containers are minimal and correct — pure stdlib HTTP servers, digest-pinned to the same `python:3.13-slim` as the real images so the supply-chain pin test passes; graceful 503s pre-cutover.
- Cutover doc honestly addresses the CFN-can't-relocate-resources problem, Memory's transitional-state gotcha, and pre-cutover backup checklist.
- `backend.yml` per-image parallel jobs with content-hash skip is a nice CI ergonomic win.
- Integration test exists that synthesizes the full stack including `wireCompute()` and asserts zero `Fn::ImportValue` references.

### Issues to address

#### High

**1. Same-stack SSM publish-then-read may deadlock on first deploy.** `app-api-environment.ts:118` and `inference-agentcore-construct.ts:124-261` use `ssm.StringParameter.valueForStringParameter()` to read ~60 parameters that are *now created by sibling constructs in the same stack*. `valueForStringParameter` synthesizes a `AWS::SSM::Parameter::Value<String>` CFN parameter, which CFN resolves *before* any resources are created. On a fresh first deploy, those SSM params don't exist yet → CFN fails with \"Unable to fetch parameters from parameter store.\"

The author acknowledged this exact risk for *two* refs (`app-api-environment.ts:97-110`) and threaded `memoryId` + `inferenceApiRuntimeEndpointUrl` via props — but the same reasoning applies to every other SSM read, since after Phase 7 everything is the same stack. Recommend verifying against a clean AWS account, or migrating same-stack reads to `StringParameter.fromStringParameterAttributes` (emits `{{resolve:ssm:...}}` dynamic references that resolve at resource-creation time and honor the dependency graph). The synth-time integration test won't catch this either way.

**2. \`SHARED_CONVERSATIONS_TABLE_NAME\` is wired to the wrong table.** `app-api-environment.ts:240`:
\`\`\`ts
SHARED_CONVERSATIONS_TABLE_NAME: params.ragAssistantsTableName, // TODO: should be shared-conversations SSM
\`\`\`
A `SharedConversationsConstruct` exists in the same stack and creates a real shared-conversations table, but the app-api env var points at `rag-assistants`. This is a functional bug — sharing a conversation will write to / read from the assistants table. Fix by publishing `shared/conversations-table-name` to SSM in the construct (or pass via prop). Don't ship with a `TODO` on a misrouted production var.

#### Medium

**3. Stale `BackendStack` references throughout comments and JSDoc.** After Phase 7, BackendStack no longer exists, but its name appears in dozens of comments — `platform-stack.ts:108-114` (\"BackendStack can take them as explicit construct props\"), `app-api-iam-grants.ts:67-69`, `inference-agentcore-construct.ts:55-57`, `app-api-environment.ts:97-103`. These describe a 2-stack-era design that no longer applies. Rename to `PlatformStack` or drop the cross-stack rationale entirely.

**4. Untyped stub-fills hide misuse.** `platform-stack.ts:548-550`:
\`\`\`ts
this.spaDistribution = undefined as unknown as cloudfront.IDistribution;
this.spaDistributionDomainName = undefined as unknown as string;
\`\`\`
If a caller forgets `wireSpaDistribution()`, the type system silently lies. A downstream `.distributionDomainName` access on `undefined` is a confusing runtime TypeError, not a CDK validation error. Either type as `IDistribution | undefined` and force callers to assert, or inline the SPA distribution and remove the two-phase pattern.

**5. \`platform.yml\` concurrency + path triggers.** `cancel-in-progress: false` combined with no path scoping to `infrastructure/bootstrap-assets/**` can queue multiple PlatformStack deploys that serially try to update resources in transient `CREATING` states. Consider `cancel-in-progress: true` for `develop` (false only for `main`), and add `infrastructure/bootstrap-assets/**` to the path triggers.

**6. `wireCompute` ordering correctness relies on construction order alone.** The comment at `platform-stack.ts:623-627` says ordering \"matters\" — it doesn't for correctness (token resolution + CFN dependency graph handle it), only for readability. Clarify so a future reader doesn't try to \"fix\" it by parallelizing.

#### Low

**7. Test gap on first-deploy ordering.** Add an assertion that walks all `AWS::SSM::Parameter::Value` parameters in the synthesized template and verifies each referenced SSM path is also created by a resource in the same template (or is in an explicit allowlist of externally-seeded params). Would have caught issue #1 at synth time.

**8. Per-image build jobs in `backend.yml` duplicate ~25 lines of env/auth setup three times.** Factor into a composite action.

**9. CFN parameter for `cdk.Fn.split(',', ...)` of subnet/AZ lists** at `app-api-service-construct.ts:93-95` still works but is unnecessary in a single-stack design — `vpc.privateSubnets` is available as a direct same-stack ref. Same for the ALB security group/listener/cluster imports a few lines below.

**10. `app-api-iam-grants.ts` reads several artifacts/fine-tuning SSM params at lines 244-249, 278-283.** Pass via props for consistency with the same-stack rationale already applied to `agentCoreMemoryArn` and `sagemakerExecutionRoleArn`.

### Risk assessment

- Blast radius is large but bounded by the cutover discipline — the doc forces a teardown step, reducing accidental partial-state migration.
- First-deploy failures are recoverable via the forward-fix path documented in section 4 of the cutover.
- No backend application code is changed in this PR. The bootstrap pattern means runtime behavior depends on a follow-up `backend.yml` image push — make sure that workflow has run successfully end-to-end against this branch in dev before merging.
- The wrong shared-conversations table (#2) and the SSM-deadlock risk (#1) are the only blockers; the rest are cleanup/follow-up.

### Recommendation

Approve with required follow-ups on #1 + #2. Land #3–#10 in subsequent cleanup PRs.

philmerrell:
## Code Review — Stack Architecture Simplification

Reviewed the full diff (207 files, +15,416 / −22,228, 62 commits). Below are findings, with each load-bearing issue verified against both the old (`95d51d96`) and new code rather than inferred. 🤖 Generated with [Claude Code](https://claude.com/claude-code).

### Overview

Ambitious, well-documented refactor: collapses the 9-stack (later 2-stack) CDK architecture into a **single `PlatformStack`** and introduces a **"bootstrap-container" deploy pattern** — CDK owns infra + a stable placeholder image/Lambda code, and application *code* ships out-of-band via AWS APIs (ECR push → `update-function-code` / `register-task-definition`+`update-service` / `update-agent-runtime` / `s3 sync`). Inlined resources are lifted into ~40 focused constructs, per-stack workflows are replaced by `platform.yml` (CFN) + `backend.yml` (code), and `cutover-platform-as-bootstrap.md` documents the one-time migration.

Engineering quality is high — clean construct decomposition, a genuinely good cutover doc, materially improved test gating, and graceful bootstrap handlers. But there are **three blocking issues** plus several should-fixes.

### What's done well ✅

- **Test gating is a real improvement.** Every backend image build and code-deploy job has `needs: test-backend` (full `uv run pytest tests/ -v`); `platform.yml`/`frontend-deploy.yml` gate on `test-infra`/`test-frontend`. At merge-base no workflow ran pytest as a blocking gate — this closes that gap.
- **OIDC role-assumption** with static-key fallback; `permissions: id-token: write` scoped per job; no secrets echoed; actions SHA-pinned with `# vX` comments.
- **Bootstrap handlers fail safe** — pure stdlib, health checks pass, real traffic gets 503 during the window; `services-stable` passes.
- **`integration.test.ts` asserts `Fn::ImportValue` count == 0** — a nice guard validating the single-stack collapse.
- Same-stack SSM→typed-ref purge (avoiding the first-deploy chicken-and-egg on `AWS::SSM::Parameter::Value`) is the correct fix and well-explained.

### 🔴 Blocking

**1. IAM regression: OAuth / connector grants dropped — will break `/connectors/*` and external-MCP OAuth at runtime.**
The grant extraction dropped statements the old code documented as required:
- **app-api** (`app-api-iam-grants.ts:399-410`) now grants only `GetWorkloadAccessTokenForUserId` + `GetWorkloadIdentity`. Old `app-api-stack.ts:1186-1268` additionally granted, with the comment *"Without these, `/connectors/{id}/{status,initiate,disconnect,complete}` return 503"*: `GetResourceOauth2Token`, `CompleteResourceTokenAuth`; `CreateTokenVault` + `Create/Update/Delete/Get/ListOauth2CredentialProvider`; and `secretsmanager:CreateSecret/DeleteSecret/PutSecretValue/UpdateSecret/TagResource` on `bedrock-agentcore-identity!default/oauth2/*`.
- **inference-api** (`inference-api-iam-roles.ts:223-228`) dropped `GetResourceOauth2Token` (old `inference-api-stack.ts:574-581`, comment *"called by the agent loop's tool gating"*). The oauth2-secret read is kept, but the control-plane action that mints the token is gone.

`git grep` confirms `GetResourceOauth2Token` / `CompleteResourceTokenAuth` / `*Oauth2CredentialProvider` appear **nowhere** in the new infra tree. The `app-api-iam-grants.ts` docstring claims it *"attaches all required IAM policy statements"* — inaccurate. Please confirm against current connector/OAuth usage and restore the missing grants (or document why they're intentionally gone).

**2. Bootstrap-image invariant does not hold for ECS + AgentCore Runtime under CFN.**
Both compute constructs bake the bootstrap image directly into the CFN resource with no escape hatch:
- `app-api-service-construct.ts:214` — `image: ecs.ContainerImage.fromDockerImageAsset(bootstrapImage)`
- `inference-agentcore-construct.ts:282` — `containerUri: bootstrapImage.imageUri`

The comments assert the real out-of-band image is "left untouched on subsequent deploys." That only holds when the task-def/runtime resource is *otherwise unchanged*. ECS task defs are immutable and `update-agent-runtime` is full-replacement — so when **any** property changes (a new env var, CPU/memory, authorizer config), CFN re-registers/updates with the synthesized template carrying `image = <bootstrap 503 stub>` and repoints the service/runtime to it, reverting the live image. Compounding it: `platform.yml` (paths-filtered to `infrastructure/lib/constructs/**` etc.) only runs `cdk deploy` and does **not** re-ship backend code afterward, so an infra-only change strands app-api **and** inference-api on the 503 stub until someone manually runs `backend.yml`. (The nightly pipeline chains them correctly; the standalone `platform.yml` path is the hazard. The two Lambdas are genuinely immune since `Code` is one property CFN leaves alone.)

Recommend one of: (a) `platform.yml` chains a backend re-ship after a successful `cdk deploy`; (b) a CFN escape hatch so CDK stops managing the image (EXTERNAL deployment controller / `addPropertyDeletionOverride` on `containerUri` / SSM-backed image param). Either way, correct the construct comments for the ECS + Runtime cases.

**3. Teardown can't destroy the new architecture — and silently reports success.**
`scripts/teardown/destroy.sh:61-73` uses a hardcoded `PARALLEL_STACKS` list of legacy 9-stack names + `FOUNDATION_STACK="InfrastructureStack"`, iterated with a `stack_exists`-skip. There is **no dynamic prefix discovery**. `${prefix}-PlatformStack` — the only stack this PR deploys — is not in the list, so `teardown.yml` against a post-cutover environment skips everything and **exits 0 "successfully" while deleting nothing**. The cutover doc's claim that it "deletes every stack whose name starts with `${prefix}-`" is also false given this implementation. Add `PlatformStack` (keep legacy names for migration) or switch to genuine `describe-stacks`-by-prefix discovery.

### 🟡 Should-fix

- **Docs are materially stale — and `CLAUDE.md` is unchanged.** As the agent source-of-truth this will mislead: Deploy-order still describes the 9-stack order + `CDK_ARTIFACTS_ENABLED`/`CDK_MCP_SANDBOX_ENABLED` flags (removed here); the "New CDK stack" rule points at `test/stack-dependencies.test.ts` (deleted) and `scripts/stack-<name>/` (all deleted). `infrastructure/README.md` still claims a **two-stack** architecture with a `BackendStack` + `scripts/backend/deploy.sh` (neither exists in the final state). `.github/docs/deploy/step-04-deploy.md` was not updated and is fully stale.
- **Lost security test coverage.** Deleted `app-api-stack.test.ts` / `security-best-practices.test.ts` / `inference-api-stack.test.ts` held meaningful assertions now gone: `BffCookieSigningKey` Decrypt-only (no `kms:GenerateDataKey`), "no policy has both `Action:"*"` and `Resource:"*"`", S3 encryption + public-access-block, DDB SSE. The new `app-api-iam-grants.ts` (411 lines) and `inference-api-iam-roles.ts` (291 lines) are **not exercised by any test** — the new suite is mostly `resourceCountIs`/existence checks. A policy-level test would have caught blocker #1. (The 9 local `config.test.ts` failures are environmental — stale untracked `lib/*.js`; the committed suite is green at 361/361 in clean CI.)
- **Unconditional provisioning is an un-noted behavior/cost change.** Removing the `enabled` flags means artifacts + mcp-sandbox + fine-tuning (previously default `false`) are now always created — new CloudFront distributions, DDB, S3, SageMaker IAM. Intended, but worth an explicit release-note/PR-description callout for existing deployers.
- **`restore-data.yml` `setup-uv` has no `version:` pin** — resolves latest uv at runtime, violating the exact-pin convention (`backend.yml`/nightly pin `0.7.12`).
- **Cross-workflow concurrency overlap**: `backend.yml`, `platform.yml`, and the nightly pipeline use disjoint concurrency groups, so a push deploy and a nightly run could both `update-service`/`update-agent-runtime` the same service. Low probability; consider an environment-keyed deploy lock.

### 🟢 Nits

- `compute-content-hash.sh:94` expands a possibly-empty array under `set -u`; on macOS bash 3.2 the no-`--manifest` (artifact-render) path errors. CI (bash 5) is fine; use `"${MANIFESTS[@]+...}"`.
- `scripts/frontend/deploy.sh` references `${AWS_REGION:?}` but `load-env.sh` exports only `CDK_AWS_REGION`; CI sets it at job level but local runs fail. Mirror `export AWS_REGION="${CDK_AWS_REGION}"`.
- `frontend-deploy.yml` `build` job isn't gated on `test-frontend` (only `deploy` is) — wasted CPU, not a safety hole.
- Stale `config.artifacts.enabled` doc-comments in `artifact-render-token-secret-construct.ts:25` and `spa-distribution-construct.ts:48`; five vacuous no-assertion tests in `config.test.ts`'s "Boolean Parsing" block; ~10 "stays in BackendStack" comments in `platform-stack.ts` the final collapse made false.
- `deploy-runtime-image-if-changed.sh` rebuilds the `update-agent-runtime` payload from a hardcoded allow-list — a maintenance hazard if the control API adds required fields.

### Recommendation

Strong architecture, not yet mergeable as-is. Blockers #1 (OAuth grants) and #3 (teardown) are fast fixes; #2 (image revert) needs an explicit design answer plus comment corrections. The doc drift and a regression-guard IAM test should ideally land in the same PR since they're part of the contract this refactor changes.

PR #402, opened 2026-05-28
fix(file-upload): fix duplicate document name error

### LINKED ISSUE #400: Bug: duplicate document name error on file upload (misclassified as size error)
## Summary

User reported being unable to upload a 100kb file in a conversation, receiving an error message about the file being too large — even though 100kb is well within limits.

## Root Cause

Two separate bugs working together:

**1. Overly broad error classifier (deployed)**

The old `_format_force_stop_message` condition in `stream_processor.py` matched on `validationexception + document`:

```python
if "maximum document size" in reason_lower or (
    "validationexception" in reason_lower and "document" in reason_lower
):
```

The actual Bedrock error is:

```
An error occurred (ValidationException) when calling the ConverseStream operation: Messages can't contain duplicate document names.
```

This string contains both `validationexception` and `document`, so it was **misclassified as a size error** and rendered the misleading "file is too large / enable Spreadsheet Analysis" message. The raw AWS text leaked into the `detail` field, which is why the user saw both messages.

**2. Document blocks not stripped from conversation history**

`_truncate_tool_contents` in `turn_based_session_manager.py` replaced image blocks with text placeholders during compaction, but left document blocks (with full bytes and names) intact in history. When a user re-attaches a file with the same name in a later turn (or returns to the same session URL), Bedrock sees two document blocks with the same sanitized name and rejects the request.

This is why it appears on a "first turn" — the session UUID is stored in the URL. Returning to the same session after a prior visit that included a file upload means the document block from that visit is still in AgentCore Memory history.

## User-Visible Symptom

```
⚠️ Something went wrong.⚠️ One of the attached files is too large for the model to read directly.

Bedrock limits inline documents to 4.5 MB of internal content, and spreadsheets
(especially XLSX) often expand past that when parsed. To work with large data files,
enable Spreadsheet Analysis in the Tools section of the settings panel (gear icon
next to the message input) and re-attach the file — the tool runs pandas on the
real file and isn't limited by the document size cap.

For large PDFs or docs, split them into smaller sections or convert to plain text.

An error occurred (ValidationException) when calling the ConverseStream operation:
Messages can't contain duplicate document names. Rename the document and retry your request.

Please try again.
```

File was 100kb. Error had nothing to do with file size.

## Fixes (in accompanying PR)

- **`stream_processor.py`** — narrow size-limit classifier to require explicit size markers (`too large`, `exceeds`, `document size`); add specific `duplicate document name` branch above it
- **`errors.py`** — add override in `build_conversational_error_event` for the raw-exception path (Path B) so the clean message is shown regardless of which error code path is taken
- **`turn_based_session_manager.py`** — strip document blocks from history during compaction (same treatment as images), preventing the underlying duplicate-name condition from occurring
- **`prompt_builder.py`** — deduplicate document names within a single turn using a counter suffix (`_2`, `_3`) as belt-and-suspenders
- **`routes.py`** — deduplicate files across the `files` + `file_upload_ids` merge paths before partitioning

DerrickF:
Fixes #400 — duplicate document name error on file upload (misclassified as size error)

What was happening
Users reported being unable to upload a 100kb file, seeing an error about the file being too large. The file size had nothing to do with it.

The actual Bedrock error was ValidationException: Messages can't contain duplicate document names — triggered when a user returned to an existing session and re-attached a file with the same name as one from a prior visit. The error was then misclassified by an overly broad condition in _format_force_stop_message:

# old — matched any ValidationException that mentioned "document"
if "maximum document size" in reason_lower or (
    "validationexception" in reason_lower and "document" in reason_lower
):
This matched the duplicate-name error and rendered the wrong "file too large / enable Spreadsheet Analysis" message. The raw AWS exception text leaked into the detail field, so users saw both messages concatenated.

Root cause
Document content blocks (with full inline bytes) were being persisted into AgentCore Memory and surviving session restores indefinitely. Images were already stripped to text placeholders during compaction, but documents were not. On a return visit to the same session, the SDK reloaded the prior document block from memory, and attaching a same-named file produced two document blocks with the same sanitized name — which Bedrock rejects.

Changes
turn_based_session_manager.py — new _strip_document_bytes() method replaces document content blocks with text placeholders ([Document placeholder: name=..., format=..., size=...]) on every session restore, unconditionally. This mirrors the existing image placeholder behavior but is intentionally decoupled from the compaction flag — it's a correctness fix, not an optimization, and should run regardless of whether AGENTCORE_MEMORY_COMPACTION_ENABLED is set.

stream_processor.py — narrowed the size-limit classifier to require explicit size markers (too large, exceeds, document size); added a specific duplicate document name branch above it so the correct actionable message is shown.

errors.py — added the same duplicate-name override in build_conversational_error_event to cover the raw-exception path (when Strands surfaces the error as a Python exception rather than a force_stop event).

prompt_builder.py — deduplicates document names within a single turn using a counter suffix (_2, _3) as belt-and-suspenders for the case where two files in one message sanitize to the same Bedrock name.

routes.py — deduplicates files across the files + file_upload_ids merge paths before partitioning, guarding against the same file arriving via both channels.

Testing
All 802 agent/session/streaming tests and 917 shared/API tests pass.


DerrickF:
Closing in favour of a clean branch based on develop — no merge conflicts.

PR #424, opened 2026-06-02
fix(restore): wire restored IdPs into Cognito app client SupportedIdentityProviders

colinmxs:
## Symptom

After a successful restore cycle, Cognito's hosted UI showed:

> An error was encountered with the requested page. Login option is not available. Please try another one.

## Root cause

CDK provisions the BFF app client during `platform.yml` before any IdPs exist. It creates the client with `SupportedIdentityProviders: ["COGNITO"]`. `restore.py` then restores the IdP (`ms-entra-id`) but only "noted" the app client — it never called `update_user_pool_client`, so the provider list stayed `["COGNITO"]` and the hosted UI had nothing to show.

## Fix

After creating IdPs, the app-client section now:
1. Finds the live CDK-managed client by name via `list_user_pool_clients`
2. Reads its current config via `describe_user_pool_client` (non-destructive)
3. Calls `update_user_pool_client` with the merged provider list
4. Skips if the provider is already in the list (idempotent on re-runs)

## Immediate workaround (already applied)

```bash
aws --profile dev-ai --region us-west-2 cognito-idp update-user-pool-client \
  --user-pool-id us-west-2_4plbXIzA2 \
  --client-id 5r82k1v0th46h9ug1jm0mbt26l \
  --supported-identity-providers "COGNITO" "ms-entra-id" ...
```

Login is working again. This PR makes it automatic on every future restore.

## Tests

2 new pytest cases:
- `test_cognito_restore_wires_idps_into_app_client` — verifies `update_user_pool_client` is called with the restored IdP after a first-time restore
- `test_cognito_restore_skips_idp_wiring_when_already_present` — verifies no update is made on re-run when the IdP is already in the list

`uv run pytest test_restore.py -q` — **35 / 35 pass**.

## Deploy

Script-only change. No AWS re-deploy needed. Just merge and the next `restore-data.yml` run will wire the IdPs automatically.

### DIFF
--- a/scripts/restore-data/restore.py
+++ b/scripts/restore-data/restore.py
@@ -482,6 +482,7 @@ def restore_cognito(ctx: RestoreContext) -> list[dict]:
     # this restore code iterated the top-level dict directly, which
     # yielded the dict KEYS (strings) and triggered
     # `'str' object has no attribute 'get'`. Read the wrapped list.
+    idps: list[dict] = []  # populated below; referenced by app-client wiring
     try:
         idp_obj = s3.get_object(Bucket=ctx.backup_bucket, Key=f"{root}cognito/identity-providers.json")
         idp_blob = json.loads(idp_obj["Body"].read())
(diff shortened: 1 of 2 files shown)


colinmxs:
Superseded by #425, which bundles this fix together with the federated-user safe-username fix needed to actually unblock IdP login. Together they fix the full chain: "Login option not available" → "User already exists with provider user id" → working IdP login. Recommend closing this PR and merging #425 instead.

PR #467, opened 2026-06-10
feat(skills): runtime — DB-backed load + RBAC + skills_hash + local binding (PR-6a)

philmerrell:
## PR-6a — Skills runtime: DB-backed load + RBAC + cache key + binding (opt-in)

First slice of the runtime wiring (spec §0.5 PR-6). Makes admin-authored skills actually load for the right users, fold their local tools behind the meta-tools, and cache correctly — **opt-in via `agent_type="skill"`, so the default `"chat"` path is unchanged (zero behavior change)**.

PR-6 was too big for one PR; this is the foundation. Deferred to follow-ups: **PR-6b** = folding gateway/external MCP tools behind the meta-tools (genuinely hard — they materialize as client objects, not individual callables) + the read-reference-file disclosure level + an example-skill seed; **PR-7** = flip the default.

### What's included
- **DB-backed `SkillRegistry` source** (`skill_registry.py`): `load_records()` populates the registry from `SkillCatalogRepository` records (instructions inline, `bound_tool_ids`, `compose`, `resources` manifest). `load_instructions()` is DB-aware (inline body, no file). Adds `all_bound_tool_ids()` and `bind_catalog_tools(catalog_id → live object)` — the cross-source-aware analog of the local-only `_skill_name` binding. The file scan stays as the dev/legacy path, unchanged.
- **Per-agent meta-tools** (`skill_tools.py`): new `make_skill_tools(registry)` returns a `(dispatcher, executor)` pair bound to one registry by closure. **Fixes a latent concurrency bug**: the module-global `_registry` was safe only because every file registry was identical; with per-user (RBAC-filtered) skills, concurrent invocations would read whichever registry was built last. The module-level functions + `set_dispatcher_registry` remain for the dev path.
- **`SkillAgent` rework** (`skill_agent.py`): DB discovery + **`enabled_tools` augmentation** before `super().__init__` (your call: a granted skill carries authorization — its bound tools become available, all sources) → materialize → bind/fold **local** tools behind the meta-tools; gateway/external bound tools are available to the model but not folded yet (logged, PR-6b). Degrades to plain `ChatAgent` at zero skills. Restores the original `enabled_tools` in the construction snapshot so a paused-then-resumed skill turn hashes back to the same cache slot.
- **RBAC threading + `skills_hash`** (`chat/service.py`, `chat/routes.py`): the route resolves the user's accessible skills via the shared `AppRoleService.get_accessible_skills` (`"*"` → `get_all_skill_ids()`) — import-boundary-safe, the same path the model-access check uses — and threads them into every `get_agent` call (so the app-tool-call / resume paths share one cache key). `get_agent` adds a `skills_hash` (digest of accessible ids + their `updated_at`, via `skills/freshness`) to `_create_cache_key`, so an edit to a granted skill or a role-grant change invalidates the cached SkillAgent. Empty for the chat path → key unchanged.

### Verification
- `cd backend && uv run python -m pytest tests/ -v` — full suite green (the only correctness gate).
- New/updated tests: registry DB source + `bind_catalog_tools` + composite aggregation; `make_skill_tools` per-agent isolation (no global bleed); `skills_hash` separates cache slots + chat path unaffected; `_fetch_skill_records` ACTIVE-filtering + error degradation. Existing file-based skill tests unchanged and passing.
- Import boundaries clean (`agents`/`inference_api` reach skills/RBAC only via `apis.shared`).

### Notes
- Opt-in and dark: nothing flips `agent_type` to `"skill"` yet, so production behavior is unchanged. A live end-to-end walk (spec §0.6, against a real seeded skill) lands with the PR-6b seed.
- Gateway/external **availability** works now (augmentation); their **folding** behind the meta-tools is the PR-6b design.

Spec: docs/specs/admin-skills-rbac-tool-binding.md §0.5 (PR-6), §8.

### DIFF
--- a/backend/src/agents/main_agent/skill_agent.py
+++ b/backend/src/agents/main_agent/skill_agent.py
@@ -4,21 +4,72 @@ Skill Agent — ChatAgent with progressive skill disclosure.
 Replaces individual skill tools with skill_dispatcher + skill_executor,
 injecting a lightweight skill catalog into the system prompt instead of
 loading all tool schemas upfront.
+
+Two skill sources:
+
+- **DB / admin-managed** (PR-6): when ``accessible_skill_ids`` is provided
+  (the caller resolved them from the user's RBAC roles), the registry loads
+  those ACTIVE skills from the catalog repository. A granted skill's bound
+  catalog tools are added to the agent's tool universe so they materialize
+  (skill-as-grant), and locally-resolved bound tools are folded behind the
+  two meta-tools. Gateway / external MCP bound tools are surfaced as live
+  clients but not yet folded (PR-6b).
+- **File / dev** (legacy): when ``accessible_skill_ids`` is None, the registry
+  scans ``definitions/*/SKILL.md`` and binds local ``@skill``-decorated tools
+  by their ``_skill_name`` stamp, exactly as before — unchanged behavior.
+
+When zero skills are available the agent degrades to plain ``ChatAgent``.
 """
 
 import logging
-from typing import List, Optional
+from typing import Any, List, Optional
 
 from agents.main_agent.chat_agent import ChatAgent
 from agents.main_agent.core import AgentFactory
-from agents.main_agent.skills import (
-    SkillRegistry,
-    skill_dispatcher,
-    skill_executor,
-    set_dispatcher_registry,
-)
+from agents.main_agent.skills import SkillRegistry, make_skill_tools
 
 logger = logging.getLogger(__name__)
+
+
+def _is_active_status(status: Any) -> bool:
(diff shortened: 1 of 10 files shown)


philmerrell:
Full backend suite: **3667 passed, 3 skipped, 0 failed** on a clean re-run. (An earlier run showed 1 failure in `tests/routes/test_pbt_auth_sweep.py::TestNonAdminRoleRejection::test_non_admin_roles_get_403` — a Hypothesis `max_examples=100` randomized-HTTP property test in the admin-auth domain this PR doesn't touch; it passes in isolation (209/209 in the file) and on the full-suite re-run, i.e. a pre-existing flake under CPU contention, not a regression.)

PR #515, opened 2026-06-26
feat(export-targets): save_conversation agent tool (PR-6, final)

philmerrell:
## Summary

Final PR of the conversation-export feature. Adds the **conversational surface** deferred from v1: a `save_conversation` agent tool so "save this chat to my Drive" works mid-conversation, riding the existing `oauth_required` SSE consent gate.

The spec called this "a thin add" — it isn't, because of the import boundary: `agents/` may not import `apis.app_api`, where the export core lived. So the bulk of this PR relocates the export core into `apis.shared` (per CLAUDE.md's "needed by more than one consumer → shared" rule), then adds a small context-bound tool that reuses it.

## Relocate export core → `apis/shared/export_targets/`

Moved (history-preserving renames): `models`, `adapter`, `registry`, `render`, `adapters/google_drive`. Plus:
- New `transcript.py` — `collect_transcript` lifted out of the route so the **tool and endpoint render an identical transcript** (single source of truth).
- New `service.py` — the boundary-free helpers (`connector_visible_to_user`, `resolve_export_target_token`).

`apis/app_api/export_targets/` keeps `routes.py` + a thin `service.py` holding the FastAPI `HTTPException` mapping (`resolve_export_target`, `require_export_target_token`, `http_error_for_export_target_error`); it **re-exports** the shared helpers so existing imports are unchanged. Admin routes import the registry from shared.

## `save_conversation` tool

`agents/builtin_tools/save_conversation/` — factory `make_save_conversation_tool(session_id, user_id, provider, adapter)`, mirroring the artifact/spreadsheet context-bound tools. Pages the transcript → renders → resolves the user's token → `create_document` → persists a receipt → returns an "Open in <app>" link. Writes to the adapter's default app folder (no mid-chat folder picker). **Imports only `apis.shared`** — boundary clean.

## OAuth consent gate for a direct (non-MCP) tool

The existing `OAuthConsentHook` fires for all tool calls but only maps **MCP** tools to providers. Wired the direct tool in:
- `BaseAgent` gains `oauth_tool_providers` (tool-name → provider-id); `_build_tool_use_provider_lookup` maps a direct tool's name to its connector → the hook surfaces `oauth_required` and resumes after consent. **Zero new consent machinery.**
- `SkillAgent`'s override now **composes** with the base lookup (folded-tool lookup first, then base), so the tool gates under skill mode too — covered by a regression test.

## Route wiring + self-gating

- `inference-api` chat route builds the tool + gate mapping in **both** the fresh and resume branches — resume rebuilds a fresh agent and the SDK restores the paused tool call from Memory, so the tool must be registered for the OAuth resume to complete.
- **Self-gating**: absent unless `save_conversation` is enabled *and* an export-target connector is visible to the user (multiple → first, logged). No env flag.
- `TOOL_CATALOG` entry added for discovery/RBAC (mirrors `list_spreadsheets`).

## Tests

- `save_conversation` tool: success (link + receipt), unknown format, adapter-unsupported format, consent-required fallback, adapter error.
- Route gating `_build_save_conversation_tool`: absent when disabled / no target; built with the gate mapping when enabled + visible; multiple → first.
- Gate lookup: base name→provider map, and the SkillAgent compose regression.
- Relocated the export tests to the shared import paths.
- **Full backend suite green: 4044 passed, 3 skipped.** Architecture import-boundary tests confirm `agents/` imports only `apis.shared`. (pytest is the only CI gate.)

## Manual test

Can't be previewed locally — needs real Google consent + an admin-mapped export-target connector + the model invoking the tool. Steps:

1. Map a Google connector with `export_target_adapter_id=google-drive`.
2. Enable the `save_conversation` tool for the user.
3. In a chat (connector **not** yet connected), ask: "save this conversation to my Google Drive."
4. The turn pauses with an `oauth_required` consent popup → consent → the turn resumes and the agent replies with the Drive link.
5. Reload: the persisted export receipt survives.


philmerrell:
Closing as deferred. The deterministic "Save to…" button (PR-4) + folder picker (PR-5) fully cover the export use case, so the conversational `save_conversation` agent tool is a nice-to-have rather than core. It also carries a non-trivial refactor (relocating the export core to `apis/shared/export_targets/`) that only earns its keep if the tool ships — so closing whole rather than cherry-picking. The complete design (and the gotchas found while building it) is captured for a future revisit; the branch `feature/save-conversation-agent-tool` (commit 562c581a) holds the full implementation if we want to reopen.

PR #561, opened 2026-07-06
feat(harness): Phase A — headless run entrypoint + Run now (PR-1)

philmerrell:
## What

PR-1 of the scheduled-agent-runs plan (`docs/specs/scheduled-agent-runs.md` §7): the **F1 headless run entrypoint** hardened to production shape, plus the attended **"Run now"** validation surface. Includes the proven spike commit (`spike/harness-headless-entrypoint`) and three hardening commits on top.

### The F1 primitive (`apis/shared/harness/`)
- `run_agent_headless(user_id, prompt, config) → RunResult` — per-owner Cognito mint → runtime `/invocations` → server-side SSE drain → governance floor → result materialized as a titled session. A2A-ready seam (`on_event`).
- **Headless-grant record** (`grants.py`, punch-list #1) replaces the spike's BFF-table Scan: create-on-enable from an attended session, lookup via the new sparse `HeadlessGrantUserIndex` GSI, revocation deletes the stored credential. **Login-recency policy: 30 days** (grant TTL anchored to the login that issued the pinned refresh token — the Cognito refresh-token validity ceiling), env-tunable via `HEADLESS_GRANT_MAX_AGE_DAYS`.
- `CognitoRefreshBearerAuth` is now grant-backed and **rotation-aware** (punch-list #5): a rotated refresh token is persisted back onto the grant before the mint returns.
- `build_invocations_url` deduped into the harness as the single canonical resolver; the chat proxy imports it (punch-list #4).
- `enabled_tools=None` = "all RBAC-allowed, like an attended turn" — documented, with the schedule-must-snapshot rule noted for Phase B (punch-list #7).

### Governance floor (F6a — locked decision: audit-only + wired seams)
- `on_run_start` audit stays **fail-closed** (no audit record → no run; tested: no bearer mint, no HTTP).
- `check_input` / `classify_output` remain documented **no-op seams**; Bedrock Guardrails / PII-classification are gated to the scheduled phase.

### "Run now" surface (`apis/app_api/runs/`)
- `POST /runs/now`, `GET /runs/grant`, `DELETE /runs/grant` — cookie auth (`get_current_user_from_session`) per house rule.
- Gated by **both** controls (spec §6): `SCHEDULED_RUNS_ENABLED` kill switch (default ON; only literal `"false"` disables — empty workflow var safe) and the new **`scheduled-runs` RBAC capability**, resolved through the mature tools grant axis (`apis/shared/rbac/capabilities.py`). GA = grant the id to the `default` role.
- Mint failures are 409, never 401 (no SPA login-redirect bounce).

### CDK
- `scheduledRuns.enabled` in `config.ts` (kbSync ternary copied exactly) → `SCHEDULED_RUNS_ENABLED` on app-api → `CDK_SCHEDULED_RUNS_ENABLED` in `platform.yml`.
- Sparse `HeadlessGrantUserIndex` GSI on the BFF sessions table (existing IAM grant already covers `index/*`).

## Out of scope (per plan)
Scheduler/EventBridge/dispatcher (Phase B), schedule data model, SPA, Guardrails/PII implementation, connector email, cohort role-management UI, dev-ai vault-3LO catalog gap.

## Tests
- Backend: **4208 passed, 3 skipped** (full suite), incl. new grant-lifecycle, grant-backed-mint, audit-fail-closed, runner-outcome (MockTransport), and route-gating tests; architecture import-boundary tests pass.
- Infrastructure: `tsc --noEmit` clean; **jest 429 passed** incl. new flag-matrix + GSI-shape tests.
- No new packages; exact pins untouched.

### DIFF
--- a/.github/workflows/platform.yml
+++ b/.github/workflows/platform.yml
@@ -96,6 +96,12 @@ jobs:
       # `CDK_KB_SYNC_ENABLED` variable to "false" in an environment only to
       # dark-stop the feature there.
       CDK_KB_SYNC_ENABLED: ${{ vars.CDK_KB_SYNC_ENABLED }}
+      # Scheduled runs (headless "Run now" + grant routes, and the Phase-B
+      # scheduler when it lands). Default ON with a kill switch: unset
+      # resolves to empty string, which config.ts treats as the default
+      # (on). Set the `CDK_SCHEDULED_RUNS_ENABLED` variable to "false" in
+      # an environment only to dark-stop the feature there.
+      CDK_SCHEDULED_RUNS_ENABLED: ${{ vars.CDK_SCHEDULED_RUNS_ENABLED }}
       # Secrets
       AWS_ROLE_ARN: ${{ secrets.AWS_ROLE_ARN }}
       AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
--- a/backend/scripts/spike_headless_run.py
+++ b/backend/scripts/spike_headless_run.py
@@ -1,0 +1,211 @@
+"""Dev-ai driver for the headless run-entrypoint spike (F1).
+
+Runs `apis.shared.harness.run_agent_headless` from a laptop against the
+deployed dev-ai AgentCore Runtime — through the runtime gateway, exactly the
+path a scheduler worker would take. See
+docs/specs/harness-entrypoint-spike-findings.md.
+
+Usage (requires an authenticated AWS profile for the dev-ai account):
+
+    cd backend
+    AWS_PROFILE=dev-ai uv run python scripts/spike_headless_run.py \
+        --user-id <cognito-sub> \
+        --prompt "Find 3-credit undergraduate communication classes" \
+        --tools class_search
+
+    # Negative probes for the record (gateway auth evidence):
+    AWS_PROFILE=dev-ai uv run python scripts/spike_headless_run.py \
+        --user-id <cognito-sub> --probe-workload-token --probe-sigv4
+
+The script resolves all names from SSM / naming conventions for --prefix
+(default dev-boisestateai-v2), exports the env vars the shared harness
+expects, runs the turn, then reads back the session-metadata row and the
+RUN# audit record as delivery proof.
+"""
+
+from __future__ import annotations
+
+import argparse
+import asyncio
+import json
+import logging
+import os
+import sys
+import urllib.parse
+import uuid
+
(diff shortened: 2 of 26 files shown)


philmerrell:
## Review (orchestrator pass — security surface read in full)

**Verdict: strong PR, faithful to the locked decisions. No Phase-A blockers.** grants/auth/route/governance are well-designed — revoke REMOVEs the credential in the same write, the mint never shortcuts to the caller's live token, gating order is correct (cookie → flag 404 → capability 403), 409-not-401 is right, and the audit floor is genuinely fail-closed.

### 1. Substantive — flag for Phase B now (not a Phase-A blocker)
`classify_output` is documented as the PII/FERPA checkpoint that runs **"before delivery"** — but per the findings, the runtime `/invocations` turn **already materializes the session + messages + title during the turn**, and the harness only adds an idempotent ensure + title override afterward. So by the time `classify_output` runs post-turn, the unclassified output is **already persisted and visible in the user's session list** — the checkpoint can't actually *prevent* delivery.

For attended "Run now" (no-op seam) this is fine. But the locked decision says **scheduled runs require the PII checkpoint filled before they ship**, and as structured that checkpoint would be too late. Phase B needs the classification to either run **in-loop (inside the turn, before the runtime persists)** or become a **post-hoc scrub/delete** of the already-persisted message — not a pre-delivery gate. Worth resolving in the scheduled-runs design before building on this seam. (No code change needed in this PR — just don't let the seam's "before delivery" docstring imply a guarantee it can't keep once the scheduler arrives.)

### 2. Capability rides the tools grant axis — confirm no stray wildcards
`user_has_capability` → `can_access_tool`, so **any role with a `*` tools grant implicitly holds `scheduled-runs`**. That's intended for `system_admin`, but before the cohort relies on this gate, confirm no *other* role carries wildcard tools (a "power user" role with `*` would silently be in the beta). Author documented this and the first-class-axis upgrade path — just verify the current role set.

### 3. `request.state.bff_session` dependency — verify in the target env
Run-now's create-on-enable pins the grant from `request.state.bff_session`. Given the in-flight BFF cookie migration, confirm that middleware actually attaches the session record in the deployed env — otherwise a cohort user with a valid cookie but no attached record hits the 409 path on first Run-now. Good dev-ai integration test.

### 4. Minor — concurrent-enable race
`enable` is get-then-put with no conditional write; two simultaneous enables could create two active grants. Low risk for attended Run-now (and `revoke` clears all active), fine for Phase A — note it if enable ever moves to a hotter path.

Nice work: full suite green (4208), no new packages, and the seam design (`BearerAuthStrategy`, `on_event`) keeps the A2A/scheduler paths open.

PR #563, opened 2026-07-06
feat(schedules): B1 — schedule data model + CRUD (inert)

philmerrell:
## Summary

Phase B, item B1 of scheduled agent runs (`docs/specs/scheduled-agent-runs.md`, `docs/specs/scheduled-runs-phase-b-brief.md`). Deliberately **inert**: schedules can be created, listed, edited, paused/resumed, and deleted — nothing fires yet. The dispatcher/worker that reads `DueScheduleIndex` and calls `run_agent_headless` is B2.

- **`ScheduledPrompt` model + `ScheduledPromptService`** in `apis/shared/scheduled_prompts/`, mirroring `apis/shared/sync_policies/` (sparse "due" GSI, `list_due_schedules`/`set_schedule_state`/`rearm_schedule`, conditional writes, delete = total revocation).
- **Cadence → `next_run_at`** (`daily` / `weekday` / `weekly`) computed timezone-aware (stdlib `zoneinfo`, IANA tz) in the service — no cron strings; the future dispatcher stays a dumb "who's due ≤ now" query.
- **app-api CRUD** under `apis/app_api/schedules/` at the `/schedules` prefix: create / list / get / update (fields + pause/resume via `state`) / `POST .../pause` / `POST .../resume` / delete. Every route uses `Depends(get_current_user_from_session)` (cookie auth) + `SCHEDULED_RUNS_ENABLED` flag (404 when off) + `scheduled-runs` RBAC capability (403), reusing Phase A's exact gating pattern from `apis/app_api/runs/routes.py`.
- **`enabled_tools` snapshot at creation** (Phase A punch #7): an explicit list from the caller is trusted as-is; `None` resolves the caller's current RBAC-allowed tool set via `AppRoleService.resolve_user_permissions` and freezes it — never re-resolved lazily at fire time.
- **Runaway-guard fields** (`runsToday`, `runsTodayDate`, `maxRunsPerDay`) present in the model now so B2 needs no migration.
- **CDK**: added a sparse `DueScheduleIndex` GSI (`GSI3_PK`/`GSI3_SK`) to the existing `SessionsMetadataTable` construct (`ScheduledPrompt` rides `PK=USER#{user_id}, SK=SCHEDPROMPT#{schedule_id}` on that table, alongside the harness's `RUN#` audit records). Distinct attribute names avoid colliding with `SessionLookupIndex`'s existing `GSI_PK`/`GSI_SK`. app-api's task role already grants full CRUD + `index/*` on this table, so no IAM changes were needed.

## Test plan

- [x] `cd backend && uv run python -m pytest tests/ -v` — 4283 passed, 3 skipped, 0 failed (includes 68 new tests: cadence math incl. timezone/weekday/weekly, sparse due-index behavior, conditional re-arm, state transitions, snapshot semantics, per-user cap, ownership isolation, and all 3 gating layers)
- [x] `tests/architecture/test_import_boundaries.py` — 6/6 passed (no cross-boundary imports introduced)
- [x] `cd infrastructure && npx tsc --noEmit` — clean
- [x] `cd infrastructure && npx jest` — 429/429 passed

## Explicitly out of scope

Dispatcher/worker Lambdas + EventBridge (B2), SPA (B3), governance/PII work (B0), email delivery, any actual run execution. `ScheduledPrompt` is not wired to `run_agent_headless`.

### DIFF
--- a/backend/src/apis/app_api/main.py
+++ b/backend/src/apis/app_api/main.py
@@ -201,6 +201,7 @@ from apis.app_api.voice import router as voice_router
 from apis.app_api.user_menu_links.routes import router as user_menu_links_router
 from apis.app_api.system_prompts.routes import router as system_prompts_router
 from apis.app_api.runs.routes import router as runs_router
+from apis.app_api.schedules.routes import router as schedules_router
 
 # Include routers
 app.include_router(health_router)
@@ -235,6 +236,7 @@ app.include_router(voice_router)  # Cookie-authenticated WS proxy for Nova Sonic
 app.include_router(user_menu_links_router)  # Public read of admin-managed user-menu links
 app.include_router(system_prompts_router)   # Public read of admin-managed system prompts
 app.include_router(runs_router)  # Headless "Run now" + grant lifecycle (scheduled-runs PR-1; SCHEDULED_RUNS_ENABLED + RBAC gated at runtime)
+app.include_router(schedules_router)  # Schedule CRUD (scheduled-runs B1; inert — SCHEDULED_RUNS_ENABLED + RBAC gated, nothing fires yet)
 
 # Conditionally register fine-tuning routes
 if os.environ.get("FINE_TUNING_ENABLED", "false").lower() == "true":
--- a/backend/src/apis/app_api/schedules/__init__.py
+++ b/backend/src/apis/app_api/schedules/__init__.py
@@ -1,0 +1,1 @@
+"""Schedule management routes (scheduled agent runs — B1, inert CRUD)."""
--- a/backend/src/apis/app_api/schedules/models.py
+++ b/backend/src/apis/app_api/schedules/models.py
@@ -1,0 +1,118 @@
+"""Schedule API request/response models."""
+
+from typing import List, Literal, Optional
+
+from pydantic import BaseModel, ConfigDict, Field, model_validator
+
+from apis.shared.scheduled_prompts.models import ScheduleCadence, ScheduledPrompt, ScheduledPromptState
+
+_MAX_PROMPT_CHARS = 20_000
+
+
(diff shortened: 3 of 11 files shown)


philmerrell:
## Review (orchestrator pass — inert CRUD, lighter than the Phase A pass)

**Verdict: clean and correct, faithful to `sync_policies`. No blockers.** Verified the scheduler-critical bits:

- **Sparse due-index** — `GSI3_PK/GSI3_SK` set only when `state=="active"`; `set_schedule_state` REMOVEs them on any pause, so paused schedules are physically absent from `DueScheduleIndex`. ✅
- **Due query** — `lte(f"{now}~")` correctly bounds `{ts}#{id}` keys to `ts <= now` (fixed-width ISO + `~` sorting after `#`/digits). ✅
- **Idempotent re-arm** — conditional on `nextRunAt == :expected`; the second dispatcher of a double-tick loses and skips. ✅ (Ready for B2.)
- **Ownership isolation is structural** — every op keys by `PK=USER#{user_id}`; a user can't address another's schedule. ✅
- Snapshot `enabled_tools` at create (punch #7), per-user cap, delete = total revocation. ✅

### Minor notes (non-blocking, for B3/later)
1. `update_scheduled_prompt` skips `None` values, so there's no way to *clear* an optional field back to null (e.g. detach `assistantId`, or reset `enabledTools`). Fine for now — note it when B3 builds the edit form (it'll want an explicit "clear" affordance or a sentinel).
2. `create`/cap is get-then-put — two concurrent creates could exceed the cap by one. Benign at this surface; same pattern as the Phase A grant.

Nice: no new packages (`zoneinfo` is stdlib), DST-safe cadence math, and the `record_run_result`/`rearm`/`list_due_schedules` methods are pre-built and unit-tested so B2 needs no behavior change here.

Issue #705, opened 2026-07-20
Enh: View assistants by list or grid


kimstate:
Have the option to see the assistants as a list or a grid - similar to the feature in google drive

Screen snap of GDrive list/grid toggle: https://drive.google.com/file/d/1Z_JopsdRRcXDDj9NPWo5eloD0-9G-g-v/view?usp=drive_link

Requester: emmawelborn@boisestate.edu

DerrickF:
Closing as completed — the list/grid view toggle is live on the Agents (assistants) page. It's backed by `AgentsViewMode` in `local-settings.service.ts`, so the choice is remembered per device; both views render the same agents with the same controls (`agents.page.ts`, `setViewMode()` / `viewMode`). Confirmed clickable in the running app. Reopen if the behavior isn't what was intended.

PR #767, opened 2026-07-27
feat(agents): retheme the Agent surfaces as a store — launch card, Discover storefront, My Agents list view

philmerrell:
Three related passes over the Agent surfaces, all frontend, **no backend changes and no new endpoints**. Every field these screens render is already on `Agent`, `AgentListing` or `PinnedAgent`.

Commits are separable and reviewable in order.

---

## 1. The chat launch card replaces the assistant card

`assistants/components/assistant-card.component.ts` was the last pre-Marketplace surface, and it broke continuity on the axis that matters most: it hashed the **first letter of the name** into its own 26-entry pastel palette, while every other surface hashes `agentId` through FNV-1a into the twelve curated gradients. The same Agent was drawn two different ways either side of a tap — exactly what the note on `hashAgentId` exists to prevent. Two of those 26 (`J`, `P`) were also near-white under white text, which is an AA failure rather than a taste call.

The replacement is the store's detail page folded to chat width, in the reading order the user just tapped through: tile → name → tagline → publisher → what it can reach → what to ask.

- Tile is `app-agent-icon` at 52px, so store, My Agents, chat header and card all draw the same artwork.
- Reads the **`Agent`** shape, not `Assistant` — tagline, publisher, category and capabilities exist only there. **No new fetch**: `session.page` already loaded it to lock the chat-input pickers. Falls back to the `Assistant` shape so the card still paints before the Agent resolves, or with `/agents` disabled.
- D6 run line from a best-effort `getRunnability`, advisory exactly as on the detail page.
- Add/Added and the detail link are gated on a published listing — a private agent has no detail page and nothing to add.
- Starters stay **buttons** (unlike the detail page's read-only list): the chat has real composer prefill behind `starterSelected`.

`models/runnability.ts` gives the D6 sentence one source; the detail page now reads from it too, so the two surfaces can't drift into two phrasings of one state.

`assistants/components/assistant-card.component.ts` is deleted — no consumers remain, and the folder's README consumer map is updated. **`assistants/` itself stays**: the share dialog, KB dialogs and services still live there.

## 2. Discover reads like a storefront

- **Spotlight** — `featured[0]` gets the front door at 84px with Start chat and Add. The band is tinted by the agent's own tile gradient via `gradientFor(agentId)`, so the storefront changes character with what's on it and can never clash with the artwork sitting in it.
- **Store tile** replaces the shelf row here. Artwork to 52px, and the `+` is **always visible** — the row had it at `opacity-0 group-hover:opacity-100`, which on a touch device means the store's primary verb doesn't exist. Still no model chip, no counts, no runnability badge (D4).
- **Category chips** filter the loaded shelves; tapping the active chip clears it. Hidden while searching.
- **Your agents** becomes a rail of tiles rather than name pills.

`app-agent-listing-row` is untouched and still serves the Pinned tab.

## 3. My Agents gains a list view

A grid/list toggle in the header, remembered per device via `LocalSettingsService`; grid stays the default. Both views render the same agents with the same controls — the toggle changes density, never what's available.

That's why `app-listing-status` gained a `part` input: the list row needs the state badge inline beside the name and the reviewer's note on its own line below. It's a **split, not a filter** — no caller may render `'badge'` and drop the note. `'all'` is the default, so the grid card is unchanged.

---

## Reviewer notes

**The scrim over the spotlight gradient is load-bearing, not decoration.** The twelve palette entries were tuned to keep a white-ish *emoji* legible at tile size; two of them (amber→orange, lime→green) put white *body text* under 4.5:1 on their own. The fixed scrim makes contrast a property of the component rather than of which agent an admin happened to feature, so no future palette entry can silently break the band. A spec asserts it's always present. Please don't "let the colour through" — darken the palette instead.

**Shelves stayed a 3-up grid, deliberately not a horizontally-scrolling rail.** The rail is the app-store shape and it's wrong at this corpus size: one holding three items reads as a broken carousel, and with `GSI5_SK` on `created_at` there's no ranking to justify hiding items off-screen. "Your agents" *is* a rail — that list is genuinely unbounded and ordered by the reader.

**`featured.slice(1)` renders as "Also featured"** rather than folding into the category shelves. Those page at 12, so an older featured agent could otherwise fall off the shelf it belongs to and vanish silently.

**Not done, on purpose.** The mockups also proposed grouping My Agents by publication state and collapsing the four lifecycle buttons into a `⋯` menu. That's the arguable change and it wasn't requested — the grid card is unchanged here.

## Verification

`npx tsc --noEmit` clean. Full SPA suite green: **146 files / 1634 tests**, including new specs for the launch card, store tile, spotlight, the Discover page (spotlight/also-featured split, category filtering, search collapse), the `listing-status` part split, and view-mode persistence.

Not smoke-tested in a browser — `SKIP_AUTH=false` locally. Worth a manual pass on:

1. `/agents/discover` — spotlight Add flips the same agent's tile in the shelf below.
2. Feature a **light-gradient** agent (amber or lime tile) and confirm the spotlight text is still readable.
3. Tap `+` on a tile **on a phone** — visible without hovering. Check a role-locked pin shows the lock instead.
4. `/agents` — toggle grid/list, reload, confirm it stuck; check a `changes_requested` agent shows the reviewer's note in **both** views.
5. Dark mode and a narrow viewport throughout.


philmerrell:
#705 is addressed here

PR #781, opened 2026-07-29
fix(marketplace): require PUBLIC visibility to publish an agent

philmerrell:
## The bug

The store gates browse on `listing.state` alone; pinning gates on `visibility`. So an agent could be **published while still SHARED or PRIVATE** — a tile everyone sees and only the author can open.

Two users hit this during the dev-ai marketplace demo on 2026-07-28. `POST /agents/{id}/pin` returned a bare `Agent not found: ast-…` for a tile the store had just offered them:

```
apis.shared.assistants.service - WARNING - Access denied: user d8a13390-… (derrickfink@boisestate.edu)
attempted to access SHARED assistant ast-648cc89cb1b4 without share record
INFO: "POST /agents/ast-648cc89cb1b4/pin HTTP/1.1" 404 Not Found
```

`ast-648cc89cb1b4` was `listing.state: published` + `visibility: SHARED`, shared with exactly one person.

## Why it looked intended

`_reachability()` (from #773) treated this as advisory, and its comment said publishing a SHARED agent to a team was legitimate. It isn't: **the marketplace is public-only.** Sharing with named coworkers is a separate mechanism with its own control on the agent tile, and `AgentListing` has no audience field — the store is one global shelf. A published non-PUBLIC agent is incoherent state, not a team listing. Both that comment and the SPA's copy of it are corrected here.

## The change

- **Block submission unless PUBLIC.** Follows the existing D7 block pattern exactly: `_visibility_block_reason` is shown by `preflight_listing` and enforced by `submit_listing`, so the dialog and the transition cannot drift apart. A refusal rather than a silent widening — publication must not be a side door that changes who can reach an agent. Ordered after the memory-space block because it is the cheaper fix.
- **Re-check at approval.** `visibility` is an independent axis the author can narrow between submitting and being reviewed, so the submit-time gate says nothing about approval time. `request_changes` still works, so a reviewer can send it back.
- **Legible 403 on pin.** For an agent the store *already advertised*, existence is not a secret, so the 404 collapse withholds nothing and just contradicts the page the user is looking at. Everything else still collapses to 404, and the extra lookup is best-effort — it can never escalate a clean 404 into a 500.

`_reachability` stays, now precisely scoped: an agent published as PUBLIC and **narrowed afterwards** is the case no submit-time gate can catch.

## No SPA changes needed

The submit dialog already renders a `blockReason` as a red alert and hides the form, and block/warning are mutually exclusive in the template — so the author gets one actionable message. The pin service already surfaces the backend's `detail` verbatim, which is why the old error read `Agent not found: ast-…`.

## Testing

- 12 new tests: submit block (PRIVATE + SHARED), preflight surfacing, block ordering, approve-time re-check, `request_changes` still allowed, the 403, the still-collapsed 404, and the fail-safe fallback.
- 106 targeted tests pass. Both test fixtures now default to `PUBLIC`, since a publishable agent is the baseline and the block is the special case.
- Full backend suite: 5433 passed with 1 unrelated failure; a clean-tree run also had exactly 1 failure but a *different* test, and both pass in isolation — pre-existing order-dependent flakiness, not this change.

## Follow-up

Prevents new occurrences; does not clean up existing ones. The two affected dev-ai records are being corrected separately.

### DIFF
--- a/backend/src/apis/app_api/agent_designer/routes.py
+++ b/backend/src/apis/app_api/agent_designer/routes.py
@@ -561,17 +561,21 @@ async def agent_listing_preflight_endpoint(
 ):
     """What the submit dialog needs before the author commits (D7, owner only).
 
-    A read-only rehearsal of the submit checks: the skills publication would expose,
-    and the memory-space block if there is one. Declared before ``/listing/submit``
-    only for reading order — the paths are literal and do not collide.
+    A read-only rehearsal of the submit checks: the skills publication would expose, the
+    memory-space block if there is one, and whether the author still has to consent to
+    going public. Declared before ``/listing/submit`` only for reading order — the paths
+    are literal and do not collide.
     """
     try:
-        exposed, block_reason, reachability = await preflight_listing(agent_id, current_user)
+        exposed, block_reason, reachability, requires_public = await preflight_listing(
+            agent_id, current_user
+        )
         return ListingPreflightResponse(
             agent_id=agent_id,
             exposed_skills=exposed,
             block_reason=block_reason,
             reachability=reachability,
+            requires_public=requires_public,
         )
     except ListingError as e:
         raise HTTPException(status_code=e.status_code, detail=e.message)
--- a/backend/src/apis/app_api/agent_designer/services/listing_service.py
+++ b/backend/src/apis/app_api/agent_designer/services/listing_service.py
@@ -220,13 +220,21 @@ def _visibility_block_reason(assistant: Assistant) -> Optional[str]:
     )
 
 
-def _visibility_block(assistant: Assistant) -> None:
-    """Block submission unless the Agent is ``PUBLIC``.
+def _visibility_block(assistant: Assistant, *, consented: bool) -> None:
+    """Refuse an Agent that is not ``PUBLIC`` and whose author has not consented to it.
 
-    Deliberately a refusal rather than a silent widening: publication must not be a side
-    door that changes who can reach an Agent. The author says "make this public" in the one
-    place that means it, and the store trusts ``visibility`` rather than overwriting it.
+    Not a silent widening — publication must never be a side door that changes who can
+    reach an Agent. But the refusal alone made the *common* path a dead end: every Agent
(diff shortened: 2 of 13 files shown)


philmerrell:
## Follow-up: the block was a dead end, so the dialog now resolves it

The first commit made non-PUBLIC a hard block. That was right for the invariant and wrong for the flow: **every agent is created `PRIVATE`**, so the block fires on essentially every first submission, and the modal collects category/tagline/note but has never touched visibility — that lives on the separate agent editor page. The author was being sent to another screen and back for one decision, which is worse than the amber warning it replaced.

`2ece269` puts the control where the decision is made.

**Consent, not a side door.** The dialog shows a "Make this agent public" checkbox with copy that says what it changes *from* (`only you can open it` vs `only the people it is shared with`). `makePublic` rides the submit request and `write_listing` widens visibility **in the same write** as the listing — two writes could leave an agent listed but unreachable, which is exactly the state this gate exists to prevent. The box starts unticked, Submit is disabled until it is ticked, and the flag defaults to `false`, so a direct API caller who omits it is refused exactly as before. An already-public agent never has its visibility rewritten.

**`blockReason` and `requiresPublic` are now separate signals.** A block is "leave this dialog and go fix something" and hides the form; needing to go public is resolvable in place. Collapsing them is what created the dead end. The memory-space block is now the only true dead end left, and ticking the box does not wave it through.

**Removed `reachabilityAuthorMessage`** — it told authors to "set Visibility to Public first", which is now wrong advice. `reachability.ts` is reviewer-facing only; an agent published as PUBLIC and **narrowed afterwards** is still the case no submit-time gate can catch.

### Testing
- 7 new backend tests (consent widens in one write, already-public not rewritten, consent does not bypass the memory-space block, omitted flag still refused, preflight separates the two signals).
- New `submit-listing-dialog.component.spec.ts` — 7 tests over the consent gating, via DI tokens rather than `vi.mock` per project convention.
- SPA: 150 files / 1663 tests pass, `tsc --noEmit` clean. Backend `routes/` + `shared/` + `architecture/`: 2263 pass.

PR #890, opened 2026-08-28
fix(marketplace): stop the review flow reporting failures twice

philmerrell:
Found while verifying the review surface on dev.

## The problem

Clicking **Approve** on a submission whose author had set visibility back to Private showed the backend's message twice: inline on the page, next to the button that was just pressed, and again in the global error toast in the corner.

The toast is the worse copy of the two — further from the control that caused it, and it disappears on its own.

## The fix

Opt the review flow out via the existing `SUPPRESS_ERROR_TOAST` context token (the same one `file-source.service` and three others already use). Four calls, each of which already renders the backend's own message inline:

- the submission read
- the review diff
- the review decision (approve / request changes / decline)
- the withdrawal decision

The diff's inline message is load-bearing rather than merely nicer: the backend distinguishes *"this submission predates version snapshots, ask the author to resubmit"* from a transport failure, and the component renders that distinction. A toast flattens it into "something went wrong", which is advice the reviewer cannot act on — retrying can never work.

## Deliberately not service-wide

There is a test pinning that. `takedown` has no inline error region on the Listings page, so the toast is its only surface — silencing it would turn a visible failure into a silent one, which is strictly worse than the duplication this PR removes. A call earns the opt-out by having inline UI, not by living in this service. The spec asserts per call, so the list is also the record of which callers have been checked.

## Verification

`tsc --noEmit` clean; 84 marketplace-admin tests pass (5 new).

The new spec asserts the exact `HttpContext` token the interceptor reads, at the interceptor boundary.

**Not verified in a browser.** The change is not on dev yet, and `localhost:4200` came up on a first-boot "create admin account" screen rather than talking to the dev backend, so there was no route to seeing it live. The rest of the review surface *was* verified on dev — see below.

## While I was in there (all on dev, all passing)

- The test drive runs end-to-end: `POST /api/chat/stream → 200`, agent replied.
- Its session is `preview-…` and the sidebar thread count was identical before and after a full reload — nothing persists to the author's conversations, as the banner claims.
- Conversation area measures 567px in-column (was ~120px) and the expand control takes the panel from 332px to 688px wide.
- Approve on a private agent is refused, the specific message renders, the reviewer stays on the page and the button re-enables. Nothing was published.
- The decline dialog's confirm button is disabled with no reason and enables once one is typed.

### DIFF
--- a/frontend/ai.client/src/app/admin/marketplace/pages/submission-review.page.spec.ts
+++ b/frontend/ai.client/src/app/admin/marketplace/pages/submission-review.page.spec.ts
@@ -237,6 +237,26 @@ describe('SubmissionReviewPage', () => {
     expect(text(fixture)).toContain('no marketplace listing to review');
   });
 
+  it('reports a refused decision beside the buttons, not at the top of the page', async () => {
+    // The decision bar is sticky and the read above it is not, so on a submission with
+    // real instructions a message rendered at the top is off-screen at the moment the
+    // reviewer presses the button. That gap is what the global toast used to cover.
+    mockService.review.mockRejectedValue({ error: { detail: 'Visibility is now Private.' } });
+    const fixture = await render();
+    button(fixture, 'Approve').click();
+    await fixture.whenStable();
+    fixture.detectChanges();
+
+    const el = fixture.nativeElement as HTMLElement;
+    const alert = [...el.querySelectorAll('[role="alert"]')].find((a) =>
+      a.textContent?.includes('Visibility is now Private.'),
+    );
+    expect(alert).toBeTruthy();
+    // In the same sticky container as the decision buttons.
+    const bar = button(fixture, 'Approve').closest('.sticky');
+    expect(bar?.contains(alert!)).toBe(true);
+  });
+
   it('keeps the reviewer on the page when a decision fails', async () => {
     mockService.review.mockRejectedValue({ error: { detail: 'Visibility is now Private.' } });
     const fixture = await render();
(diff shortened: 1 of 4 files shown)


philmerrell:
Verified on a local stack against the running backends, and it turned up a real gap in the original change — pushed a second commit.

**The toast suppression works.** Approve refused on the private test agent: exactly one `[role="alert"]` on the page, `app-error-toast` renders no item, and the backend's specific message appears. (For contrast, the same action on dev — which does not have this branch — shows the "Invalid Request" toast in the corner as well.)

**But the inline message it leaves behind was in the wrong place.** It rendered at the top of the page, measured **565px** above the Approve button that produced it. The decision bar is sticky and the read above it is not, so on a submission with real instructions the reviewer presses Approve at the bottom of a scrolled page and the explanation is off-screen above. The test agent I used has no instructions, which is the only reason the first check looked fine.

That is the gap the toast was quietly covering. Removing the toast without moving the message would have turned a *duplicated* failure into a *silent* one — strictly worse than what I set out to fix.

So `554c0e71` splits the two error regions. They are read at different moments and from different scroll positions: a load failure is the first thing on an otherwise empty page and stays at the top; a refused decision now renders inside the sticky bar, directly above the buttons. Re-verified live — the error sits 17px above the bar, inside it, still no toast, still exactly one alert.

Also correcting my PR description above: where it says the message appears "next to the button that was just pressed", that was true only after this second commit.

1942 frontend tests pass (2 new here); `tsc --noEmit` clean.

PR #920, opened 2026-09-03
feat(artifacts): artifact sharing PR-2 — owner UI (Share button + share modal)

philmerrell:
PR-2 of the artifact sharing spec (`docs/specs/artifact-sharing.md`). **Owner-side UI only** — recipient route/page is PR-3, session-delete cascade is PR-4.

> **Depends on [#919](https://github.com/Boise-State-Development/agentcore-public-stack/pull/919)** (PR-1, backend). Branched off `develop` rather than stacked on #919, so there is no stacked-PR merge hazard — but the Share button calls endpoints #919 adds, so **merge #919 first**. Frontend CI is independent and green either way.

## What this adds

**`ArtifactShareService`** (`session/services/artifacts/artifact-share.service.ts`) — owner CRUD (create / list / patch / revoke) plus the two recipient reads PR-1 ships: `getSharedArtifact` and `mintSharedRenderToken`. The shared mint returns the same `{url, expires_at}` shape as the owner endpoint, so the panel's iframe and `ArtifactDownloadService`'s hidden-iframe `?download=1` trick work against either path unchanged. (`GET /shared-artifacts/{id}/content` is deliberately absent — that endpoint is PR-3's backend half.)

**`ArtifactShareModalComponent`** — access-level radio, email chips, clipboard copy, and a manage/revoke list of the artifact's existing links. Create and manage live in one dialog because an artifact has no equivalent of the sessions list's manage-shares surface.

**Share button** beside Download on the artifact card and in the panel header.

## Two conventions worth calling out

**The dialog does not inherit its source's styling.** The spec says to adapt `share-modal.component.ts`, and the frontend `CLAUDE.md` says an adapted dialog copies an older one's *structure* (backdrop + centred panel + `DialogRef` wiring) and **not** its pre-redesign tokens. So this takes the conversation modal's shape and the canonical `add-curated-model-dialog` styling: `rounded-2xl`, `text-sm/6`, `bg-blue-600`, bordered panel, bordered action footer. It also honours the result contract — `undefined` from `closed` means cancelled, a concrete `ArtifactShare[]` means something was committed. Escape, backdrop and Cancel all converge on the same handler; since create and revoke commit immediately, "cancelled" can only mean *nothing was committed*.

**Labels vs. tooltips.** The card's two actions keep visible text labels (matching the existing Download button), so the visible label is contained in the accessible name per WCAG 2.5.3 and no tooltip is needed. The panel's Share button is icon-only, so it carries `[appTooltip]` per the accessibility rule in the root `CLAUDE.md`. Its icon-only siblings (Download, Close) predate that rule and are left alone — retrofitting them is out of scope here.

## The behaviour the tests are actually guarding

Both entry points share **the version the user is looking at, never HEAD**. The card renders one row per version, and the panel's ref follows its version menu — so `share()` reads the version off the same ref that drives what is on screen. A share that followed HEAD would silently change what a recipient sees, which is exactly the moving-pointer trap the spec rules out, so `artifact-card.component.spec.ts` asserts the dialog is opened with `version: 3` for a v3 card rather than trusting it.

The modal spec covers: create pinned to the opened version; owner-plus-added-emails on a limited share; the "always shows version N, later versions aren't included" copy; the just-created link not being listed twice; revoke dropping the link *and* retiring its result panel so the dialog can't offer a dead link to copy; absolute-URL building from the server's relative route; the clipboard-denied fallback; 404/403/503 error mapping; and the `undefined`-means-cancelled close contract.

## No feature flag

The card and panel only render when artifacts are enabled, which is the same `ARTIFACTS_RENDER_TOKEN_SECRET_ARN` signal that mounts the share routes in `main.py`. If you can see an artifact card, the share endpoints exist.

## Verification

- **Full SPA suite: 177 files, 2032 tests, all passing** (27 new). DI token overrides throughout, not `vi.mock`.
- `ng build` clean. The bundle-budget warning it prints is pre-existing (2.11 MB over a 5 MB budget; this PR adds ~25 KB raw).
- `src/version.ts` not stamped — nothing extraneous in the diff.
- No browser clickthrough: `SKIP_AUTH` is not set in `backend/src/.env`, so per the local convention this ships with a manual test script instead (in the review thread below).

One thing worth an eyeball at mobile width: the card now carries two labelled buttons in its last grid column. The title truncates gracefully, but the density is a judgement call.

### DIFF
--- a/frontend/ai.client/src/app/session/components/message-list/components/artifact/artifact-card.component.spec.ts
+++ b/frontend/ai.client/src/app/session/components/message-list/components/artifact/artifact-card.component.spec.ts
@@ -1,0 +1,130 @@
+import { ComponentFixture, TestBed } from '@angular/core/testing';
+import { describe, it, expect, beforeEach, vi } from 'vitest';
+import { Dialog } from '@angular/cdk/dialog';
+import { signal } from '@angular/core';
+import { ArtifactCardComponent } from './artifact-card.component';
+import { ArtifactShareModalComponent } from './artifact-share-modal.component';
+import { ArtifactStateService } from '../../../../services/artifacts/artifact-state.service';
+import { ArtifactDownloadService } from '../../../../services/artifacts/artifact-download.service';
+import { UserService } from '../../../../../auth/user.service';
+import type { Artifact } from '../../../../services/artifacts/artifact.model';
+
+const ARTIFACT: Artifact = {
+  artifactId: 'art-1',
+  version: 3,
+  title: 'Quarterly Chart',
+  contentType: 'text/html; charset=utf-8',
+  updatedAt: '2026-09-03T00:00:00+00:00',
+};
+
+describe('ArtifactCardComponent', () => {
+  let fixture: ComponentFixture<ArtifactCardComponent>;
+  let dialog: { open: ReturnType<typeof vi.fn> };
+  let artifactState: { openArtifactPanel: ReturnType<typeof vi.fn> };
+  let download: { download: ReturnType<typeof vi.fn> };
+
+  const el = () => fixture.nativeElement as HTMLElement;
+  const button = (label: RegExp): HTMLButtonElement => {
+    const match = Array.from(el().querySelectorAll('button')).find((b) =>
+      label.test(b.getAttribute('aria-label') ?? ''),
+    );
+    if (!match) throw new Error(`no button matching ${label}`);
+    return match as HTMLButtonElement;
+  };
+
+  beforeEach(() => {
+    TestBed.resetTestingModule();
+
+    dialog = { open: vi.fn() };
+    artifactState = { openArtifactPanel: vi.fn() };
+    download = { download: vi.fn().mockResolvedValue(true) };
+
+    TestBed.configureTestingModule({
+      imports: [ArtifactCardComponent],
+      providers: [
+        { provide: Dialog, useValue: dialog },
+        { provide: ArtifactStateService, useValue: artifactState },
+        { provide: ArtifactDownloadService, useValue: download },
+        {
+          provide: UserService,
+          useValue: { currentUser: signal({ email: 'owner@example.com' }) },
+        },
+      ],
+    });
+
+    fixture = TestBed.createComponent(ArtifactCardComponent);
+    fixture.componentRef.setInput('artifact', ARTIFACT);
+    fixture.detectChanges();
+  });
+
+  it('renders Share alongside Download', () => {
+    expect(el().textContent).toContain('Share');
+    expect(el().textContent).toContain('Download');
+  });
+
+  it('gives both actions a visible label inside their accessible name', () => {
+    // WCAG 2.5.3: the visible label must be contained in the accessible
+    // name. Both card actions are labelled, so neither needs a tooltip.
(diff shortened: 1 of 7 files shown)


philmerrell:
## Manual test script

`SKIP_AUTH` isn't set in `backend/src/.env`, so I didn't drive the browser. Here's the clickthrough.

**Prerequisite:** [#919](https://github.com/Boise-State-Development/agentcore-public-stack/pull/919) merged and the backend running with `ARTIFACTS_RENDER_TOKEN_SECRET_ARN`, `DYNAMODB_ARTIFACTS_TABLE_NAME`, `S3_ARTIFACTS_BUCKET_NAME` and `ARTIFACTS_ORIGIN` set. Local :4200 talks to the dev backend, so dev needs the PR-1 code deployed.

```bash
cd frontend/ai.client && npm run start
```

### 1. Share from the card
1. Open a chat and ask for an artifact ("make me an HTML bar chart of anything").
2. On the artifact card, confirm **Share** sits beside **Download**.
3. Click **Share**. The dialog header should read the artifact title and **· version 1**.
4. Leave **Public link** selected, click **Create share link**.
5. Expect a green panel: "Artifact shared", "This link **always shows version 1**. Later versions aren't included", and a copyable URL ending `/shared-artifact/<uuid>`.
6. Click **Copy link** — the button should flip to **Copied** for ~2s.

### 2. The version pin (the thing most worth checking)
7. Close the dialog. Ask the model to update the same artifact, so a **v2** card appears.
8. Click **Share** on the **v1** card → header says **version 1**. Close.
9. Click **Share** on the **v2** card → header says **version 2**, and **Existing links** lists the v1 link with "Version 1".
10. Create a v2 link. Both should now be listed with their own versions.

### 3. Share from the panel
11. Open the artifact (click the card body) to dock the panel.
12. Confirm a share icon in the header beside download; hover it → tooltip "Share version N".
13. With multiple versions, switch version in the header menu, then click Share — the dialog's version must follow the menu.

### 4. Limited share
14. Open Share, pick **Limited share**. Your own email should appear as a non-removable `(you)` chip.
15. Add `someone@boisestate.edu`, press Enter. Chip appears; the X removes it.
16. Try adding a malformed address, your own address, and a duplicate — all three should be silently ignored.
17. Create the link. **Existing links** should show a lock icon and "2 people".

### 5. Revoke
18. In **Existing links**, click **Revoke** on a link → it disappears from the list.
19. Revoke the link shown in the green result panel → the panel should disappear too (no dead link left to copy).

### 6. Regression — the card's other actions
20. Click the card body → the panel still opens (Share must not have stolen the primary action).
21. Click **Download** → the file still saves.

### 7. Density check
22. Narrow the window to ~375px (or device-emulate an iPhone). The card now has two labelled buttons in its last column — confirm the title truncates cleanly and the row doesn't wrap badly. If it's too tight, say so and I'll drop the labels to icons at a breakpoint (with tooltips).

### 8. Keyboard + a11y
23. Tab to the card: focus ring on the card, then Share, then Download.
24. In the dialog, Escape closes it. So does clicking the backdrop.
25. Run axe on the open dialog — expect no violations.

### What "broken" would look like
- Share dialog opens with the wrong version number → the version-pin wiring regressed.
- 404 on create → PR-1 isn't deployed to the backend you're pointed at.
- 403 on revoke → you're not the owner of that artifact.

PR #994, opened 2026-09-07
fix(mcp-apps): revive a torn-down MCP session for app-initiated calls

philmerrell:
> **Draft — stacked on #993, and not yet verified end to end.** See *Verification* below.

## Problem

An app-initiated `tools/call` fails with **502 Bad Gateway**:

```
app tools/call dispatch failed (tool=board_snapshot session=…):
the client session is not running. Ensure the agent is used within the MCP client context manager.
```

An MCP App calling any tool after its turn has ended hits this — for a board that re-reads state after each edit, that is every refresh.

## Cause

The call arrives *between* turns:

1. `routes.py` rebuilds the agent first, but with `cache_write=False` — *"Read the slot; never seed it."* — so it gets a **cached** agent.
2. Strands tore that agent's MCP client sessions down when the turn that built them ended.
3. `_resolve_client` returns the client the `UIToolCatalog` recorded at some earlier build, whose session is now dead.
4. `call_tool_sync` raises `MCPClientInitializationError` → `AppToolCallError(502)`.

It presents as intermittent because a call made while the turn is still streaming finds the session alive — which matches the note in `UIToolCatalog`'s docstring:

> The client's session stays alive for the agent's lifetime … so it is still active when a tool result arrives mid-stream.

True mid-stream; not true out of band.

## Fix

Reconnect the client for the duration of the call and leave it as found:

- A session that is **already live** belongs to an in-flight turn — used as-is, never stopped here.
- A **dead** session is started for the call and stopped afterwards.
- **Overlapping** app calls against the same client share one revived session via a refcount, so no call has the connection closed underneath it.

`_session_is_active` reads Strands' private `_is_session_active` and treats an unexpected client shape as active, so we never start a session we cannot own.

## Trade-offs, deliberately

- **The deeper fix is elsewhere.** `_resolve_client` accepts `agent` and never uses it; resolving the live client from the freshly built agent would be more correct, but it reaches into how tool providers are held and cached. This keeps the blast radius at the dispatch boundary. Happy to go the other way if you prefer.
- **Reconnect cost.** One connect/disconnect per out-of-band call. Fresh auth is a side benefit (the token provider is lazy), but it is not free.
- **The revival lock is held across `start()`**, so concurrent revivals serialize — including for different clients. Fine at app-call volume; worth a per-client lock if that changes.

## Tests

Two added:

- `test_revives_a_torn_down_client_session` — start called once, session live during the call, stopped after. **Fails without the fix** (`assert 0 == 1` on `starts`).
- `test_leaves_a_live_client_session_alone` — no start, no stop, session untouched.

The new `_SessionClient` fake raises from `start()` if already running, matching upstream, so reviving a live session fails loudly. Suite: 9 passed.

## Verification

Unit-tested only. I could **not** confirm it end to end: my local stack got into a bad state while testing (one restart collided with a still-running process on `:8001`, and the SPA then stopped bootstrapping), so the browser-level repro is outstanding. The diagnosis above is from the inference-api logs and the captured 502, not from watching the fix succeed in the UI.

Worth someone reproducing before this leaves draft: open an MCP App board, let the turn finish, then trigger an app-initiated tool call.

### DIFF
--- a/backend/src/apis/inference_api/chat/app_tool_dispatch.py
+++ b/backend/src/apis/inference_api/chat/app_tool_dispatch.py
@@ -24,7 +24,9 @@ rejected as not app-visible.
 from __future__ import annotations
 
 import asyncio
+import contextlib
 import logging
+import threading
 import uuid
 from typing import Any, Dict, List, Optional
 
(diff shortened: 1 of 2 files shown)


philmerrell:
## Verified end to end (post-merge)

The outstanding item from the PR description — a browser-level repro — is now done. Recording it here since this merged while still marked unverified.

**Setup:** local stack with this fix and #993 running; MCP App board (`google_tasks` / `show_task_board`) in a fresh conversation.

**Repro of the original failure:** open the board, let the turn finish, then trigger an app-initiated tool call (`Show completed` → `board_snapshot`).

**Result:** the call succeeds. `1 open task · 21 completed` renders, with no error toast.

```
=== proxy-call results ===
POST /mcp-apps/proxy-call HTTP/1.1" 200 OK
POST /mcp-apps/proxy-call HTTP/1.1" 200 OK
=== "client session is not running" occurrences ===
0
=== dispatch failures ===
0
```

Previously that same action produced `502 Bad Gateway` and, in the inference-api log:

```
app tools/call dispatch failed (tool=board_snapshot session=…):
the client session is not running.
```

A second confirmation fell out of the same run: the board's post-handshake fallback refresh is itself an app-initiated `board_snapshot`, and it recovered a board whose initial tool result had arrived empty — i.e. the revival path ran and worked without anyone clicking anything.

### One caveat on what this does *not* cover

The concurrent case is exercised only by unit test, not in the browser: I did not get two overlapping app calls against the same client in flight at once, so the refcount's "second caller joins the revived session" path is untested against a real MCP server. Worth keeping in mind if connection churn ever shows up here.

### Note for whoever picks up the follow-up

`_resolve_client` still takes `agent` and ignores it. Resolving the live client from the freshly built agent remains the deeper fix; this one keeps the blast radius at the dispatch boundary.

PR #1027, opened 2026-09-10
feat(kb): born-managed knowledge bases — provision on first upload

DerrickF:
Makes `MANAGED_KB_NEW_DEFAULT` real (rollout ladder step 2): a new agent's knowledge base is created on the **managed** backend the moment its **first document** is uploaded, so the author never has to find the Upgrade button.

Ships dark — the flag defaults off, and every path is inert until someone turns it on.

> **This PR was reworked in place.** The original revision reused `enroll()` to migrate an empty corpus at agent creation. That approach is replaced; its flag reader and infra wiring are kept. Commit `7bbd2635` is the original, `d5a050d3` is the rework, so the delta is reviewable on its own.

## Why the approach changed

`enroll` writes `retrievalEngine=managed` only at **promotion**, and the legacy ingestion handler skips a document only when the record **already** resolves to managed. So under the old design the very first document would be indexed on legacy, answered from legacy, and then indexed a second time on managed — two writers racing over one `DOC#` status, which is the §5.37/§5.38 failure shape this feature is supposed to be retiring.

Declaring the engine **up front** is what fixes it. Stacking that onto the first upload buys two more things:

- **Quota.** Managed is one Bedrock KB per agent against a ~10,000-per-account ceiling. Provisioning at agent creation spends that on prompt-only agents and abandoned drafts. At first upload it is bounded to agents that actually use RAG.
- **WYSIWYG.** The creation page's playground uploads to the *draft* and lets the author test on the spot. Provisioning at first upload means the engine they test **is** the engine they ship.

## The flow

| Step | Where | What |
|---|---|---|
| Trigger | `documents/routes.py` `generate_upload_url_endpoint` | Create the KB_Record, declare the engine managed, persist the `DOC#` row as `provisioning`, queue the job. Three conditional writes; milliseconds. |
| Defer | `kb_migration/ingestion_consumer.py` `handle_object` | The S3 event fires in seconds; `CreateKnowledgeBase` takes minutes. Returns a benign no-op instead of raising. |
| Provision + hand off | `kb_migration/provisioner.py` (new) | Build the KB, then ingest the deferred documents itself. |
| Steady state | unchanged | Once active, uploads are ordinary managed: legacy skips, the consumer ingests on the S3 event, the byte cap applies. |

**The defer is the load-bearing part.** Lambda's asynchronous retry is capped at **2** attempts — a hard service limit — so the old "knowledge base is not provisioned" raise would burn the whole window and dead-letter the first document of every born-managed agent, leaving bytes that are fine and a document that is invisible forever. Because the *job* triggers ingestion, correctness never depends on that window at all.

**Failure means legacy, never limbo.** Managed intent with no knowledge base is a dead end both ways: legacy skips the agent because the record says managed, managed cannot serve because there is nothing to serve from. So a provisioning failure **REMOVEs** `retrievalEngine` (back to legacy-by-absence, byte-identical to a record that never tried) and fails the waiting documents with copy the author can act on. A re-upload then takes the legacy path and simply works.

## Notable decisions

- **New migration state `born_managed`**, so the job inherits the sparse work-key queue, the worker lease, the generation fence and retry-until-terminal rather than growing a second dispatcher. Not spelled `provisioning` — that word is already the `provisioningState` value, and one word with two meanings is how the wrong one gets read.
- **`records.adopt_managed_engine`** is a new conditional write, deliberately *not* `promote_engine`, which is guarded on `migrationState=promote` and on catch-up convergence because it is a cutover. Here there is nothing to carry across. It also retires the one-time upgrade notice, since a KB that was never on legacy has nothing to be congratulated about.
- **The dispatcher gates each work state on its own flag** — `born_managed` on `NEW_DEFAULT`, migration states on `MIGRATION_ENABLED`. Ladder step 2 therefore provisions new agents and touches **nothing that already exists**. Gating born-managed on the migration flag would have made step 2 useless alone; enabling everything from either flag would have made step 2 a back door for step 3's blast radius.
- **The first document IS byte-capped at request time.** Worth calling out because the opposite looks reasonable: the ingestion reconcile *commits* the reservation (`reservedBytes -= n`), so a document that committed without reserving would drive the counter negative and corrupt the cap permanently.
- **Ingestion reuses `handle_object` outright** rather than reimplementing ingest → wait-indexed → wait-retrievable → terminal. That function is where §5.37/§5.38/§5.39 and the Requirement 12.3 S3-HEAD reconcile live; a second copy is a second place to forget them.
- **`provisioning` is non-terminal and non-retrievable** — the facade serves only `complete` — so it can never answer a question out of a knowledge base that does not exist yet.

## Kept vs removed from the original revision

**Kept:** `new_default_enabled()`, the app-api `MANAGED_KB_NEW_DEFAULT` env wiring and its infra test, the flag-read tests.
**Removed:** `maybe_enroll_new_default()` and its two finalize hooks in `assistants/routes.py`.

## Tests

**27 new** in `backend/tests/lambdas/test_kb_born_managed.py`. Four **mutation guards**, each verified by actually mutating the source and watching the named test fail:

| Mutation | Fails |
|---|---|
| drop the `new_default_enabled()` gate | `test_flag_off_writes_nothing_at_all` |
| drop the consumer DEFER branch | `test_defers_while_the_knowledge_base_is_being_provisioned` |
| drop the failure-path `rollback_engine` | `test_provisioning_failure_falls_back_to_legacy` |
| gate work states on either flag | `test_migration_states_are_not_swept_under_new_default_alone` |

**380 backend** tests green (born-managed, migration worker, ingestion consumer, document reconciler, kb_upgrade, records, work-index guard, legacy engine gate, documents/byte-cap/import/assistants routes, kb_backend boundary, Lambda image imports, env contract). **72 infra** (kb-migration + app-api-environment). `ruff` clean on changed files — the two `F541` in `assistants/routes.py` are pre-existing and verified verbatim on `develop`. `tsc --noEmit` clean.

One gap: `ng test` cannot bundle locally in this environment (unrelated `node:fs` type resolution, reproduced identically in a clean checkout), so the frontend specs — including the new `provisioning` label cases — run in CI rather than locally.

## Not done here

- **Pickup latency.** Up to one 15-minute dispatcher interval before provisioning even starts, so a first upload can read "Provisioning knowledge base…" for that long before the real 47–124 s create begins. The fix is a direct async worker invoke from the API alongside the work key (which stays as the durable anchor); it needs an ECS task-role `lambda:InvokeFunction` grant and a container env var threaded through the app-api construct, which does not belong in the same change as the flow itself. Recorded in the spec.
- **Turning the flag on** is still gated on Bedrock KB-count quota headroom and the KBs-by-engine inventory (task 14.5).

Nothing deployed. Spec: `.kiro/specs/managed-kb-migration/born-managed-provision-then-ingest.md` (as-built section records every decision above); the superseded `new-default-wiring.md` carries a banner explaining why.


DerrickF:
Added a **Successor design: Lambda durable functions** section to the spec (`7e58e1e1`, docs-only).

Context for reviewers: the orchestration in this PR — the `born_managed` work state, sparse work keys, lease, 15-minute dispatcher tick, consumer DEFER, one-document-per-invocation — exists solely because a Lambda is killed at 15 minutes and an S3 event gets 3 delivery attempts. Lambda durable functions (re:Invent 2025, Python supported) remove both constraints.

Step Functions was considered first and rejected: it would be the only state machine in the stack (a new AWS service for every fork maintainer) and it moves the saga out of Python into ASL, forfeiting `test_pbt_kb_migration_convergence.py` — the crash-at-every-step property test that already caught a double-promotion bug. Durable functions avoid both, since a durable function *is* a normal Lambda and the sequence stays in Python.

Not a follow-up ticket yet, for four recorded reasons: an existing function cannot be converted to durable, it forces qualified-ARN deploys (`backend.yml` currently pushes to `$LATEST`), replay determinism is a footgun against this codebase's read-record-then-branch style, and the SDKs are young. Recommended first move is a throwaway spike of born-managed alone, not an engine rewrite.

What survives any rewrite: the conditional-write guards and the declare-the-engine-before-the-object-lands ordering. Those are the correctness core, and they are independent of the orchestrator.

Issue #419, opened 2026-06-02
Admin-managed AgentCore Gateway target registration (protocol=mcp)


philmerrell:
## Summary

Selecting **"MCP Gateway (AgentCore)"** (`protocol=mcp`) on the admin tools form (`/admin/tools/new`) exposes **no configuration fields**. The backend accepts `protocol="mcp"` but stores no config and runs no validation — the option is a non-functional placeholder.

This issue tracks **Option B**: let an admin register an externally deployed MCP server as a **target on the centralized AgentCore Gateway** from the admin UI. The tool then flows to agents through the *existing* Gateway discovery path (`gateway_mcp_client.FilteredMCPClient.list_tools_sync()`) with **zero new agent-side runtime code**, while gaining Gateway semantic tool search, centralized SigV4 auth, and Gateway-side observability.

> Not to be confused with `mcp_external` ("MCP External Server"), which already ships and connects the agent **directly** to an external MCP server (no Gateway). This issue is specifically the Gateway-fronted topology.

**Full design:** [`docs/specs/MCP_GATEWAY_TARGET_REGISTRATION_SPEC.md`](../blob/develop/docs/specs/MCP_GATEWAY_TARGET_REGISTRATION_SPEC.md)

## ⛔ Dependency

**Blocked on PR #396** (`feature/stack-architecture-simplification`) merging to `develop`. #396 relocates the Gateway into `infrastructure/lib/constructs/gateway/agentcore-gateway-construct.ts` and app-api IAM into `infrastructure/lib/constructs/app-api/app-api-iam-grants.ts` — the exact files this work edits. Building pre-#396 = guaranteed conflicts. **Do not start until #396 lands.**

## The hard part

Creating a catalog row now has a **side effect in AWS** (a live Gateway target). Catalog ↔ AWS state must not diverge:
- Create the AWS target **first**; persist the DynamoDB row only on success.
- Update/delete must reconcile the target; delete AWS target first (404 = already gone), then the row.
- Partial-failure window: v1 keeps repair manual but **must log the orphaned identifier loudly**.

This lifecycle coupling — not the form — is the real work.

## Scope checklist

**Backend** (`apis/shared`, `app_api`)
- [ ] `MCPGatewayConfig` model + `GatewayListingMode` enum (`models.py`, mirror `MCPServerConfig`)
- [ ] `ToolDefinition.mcp_gateway_config` + `to_dynamo_item`/`from_dynamo_item`
[...]

PR #1032, opened 2026-09-11
feat: show OAuth connection state on tools

philmerrell:
> [!IMPORTANT]
> **Third in a stack — merge bottom-up.** #1030 → #1031 → this. Base here is `feature/tool-drawer-search-grouping`. **Retarget each to `develop` as the one below it merges.** #419 lost `gateway_target_service.py` to exactly this: the base merged first and the stacked commit stranded.

## The bug underneath

13 of the 31 tools in prod need the user to connect an account before they run, and nothing said so — the first sign was a turn failing mid-answer.

The cause was **a field that existed and was never filled**. `UserToolAccess` has always declared `requires_oauth_provider`, but `ToolCatalogService` never passed it when constructing the response, so every tool reported `requiresOauthProvider: null`. The SPA couldn't have shown connection state even if it wanted to. One line of backend; everything else here is surfacing it.

## What you get

- Rows draw a **`connected` / `connect`** chip.
- The detail pane shows a connection banner with a working **Connect**.

## Judgment calls worth review

**The chip is static, not a button.** It sits inside the row's body button, and a button nested in a button is invalid and swallows its own activation — the same rule the split row already follows. Connecting happens in the detail pane, which has room for it.

**Connect reuses `OAuthConsentService`** rather than reimplementing consent, so popup blocking, COOP-severed `window.opener`, and the completion broadcast are already handled. `ConnectorStatusService` listens for that broadcast and flips the chip — without it the row still reads "connect" right after a successful consent, which looks like the consent failed.

**Status is cached per provider, not per tool.** Gmail, Calendar, Drive and Tasks are four rows and one `google-*` connection each. Probed with the side-effect-free `getStatus`, never `initiateConsent` — the latter records a pending session server-side every time it hands back a URL.

**A failed probe stays `unknown` and draws no chip.** Telling someone they are disconnected because a status call timed out sends them into a consent flow they do not need.

**Turning a tool on while disconnected is still allowed.** The agent surfaces its own consent prompt mid-turn; gating the toggle here would be a second, contradictory gate that disagrees with the runtime.

## Verification — read this before approving

**Tested:** `ng test` 2596 passing across 225 files (10 new); backend `tests/apis/app_api/tools/` 51 passing (3 new). I confirmed the backend test **fails without the fix** (2 of 3 red) rather than passing vacuously. I also re-drove the drawer in the browser to confirm no regression to PR-1/PR-2 behaviour.

**Not tested, and I want to be explicit:** I could not exercise the chip end-to-end against a live connection. The app-api on `localhost:8000` is the main checkout's process, so it serves the unfixed `/tools` payload, and I could not stand a second one up without taking over a port the SPA hardcodes. **The chip rendering path has unit coverage but has never rendered against real data.**

To check it yourself: run app-api from this worktree on 8000, then open Settings → Tools. Dev has exactly three OAuth-gated tools — `canvas_faculty` → `canvas-faculty`, `google_tasks` → `google-tasks`, `salesforce_oit` → `contract-salesforce` — so Canvas for Faculty is the row to watch. (Prod has 13.)

### DIFF
--- a/backend/src/apis/app_api/tools/service.py
+++ b/backend/src/apis/app_api/tools/service.py
@@ -135,6 +135,11 @@ class ToolCatalogService:
                     category=tool.category,
                     protocol=tool.protocol,
                     status=tool.status,
+                    # UserToolAccess has always declared this field; nothing ever
+                    # passed it, so every tool reported `requiresOauthProvider:
+                    # null` and the SPA had no way to tell that 13 of the 31
+                    # tools in prod need a connection before they will work.
+                    requires_oauth_provider=tool.requires_oauth_provider,
                     granted_by=granted_by,
                     enabled_by_default=tool.enabled_by_default,
                     user_enabled=user_enabled,
--- a/backend/tests/apis/app_api/tools/test_oauth_provider_surfaced.py
+++ b/backend/tests/apis/app_api/tools/test_oauth_provider_surfaced.py
@@ -1,0 +1,84 @@
+"""`requiresOauthProvider` must reach the user-facing /tools payload.
+
+`UserToolAccess` has always declared the field, but the service never passed it,
+so every tool reported ``requiresOauthProvider: null``. The SPA therefore had no
+way to tell that 13 of the 31 tools in prod need an OAuth connection before they
+will run — the user only found out when a turn failed mid-answer.
+
+Dependencies are mocked; the logic under test is pure.
+"""
+
+from unittest.mock import AsyncMock, MagicMock
+
+import pytest
+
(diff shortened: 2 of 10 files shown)


philmerrell:
Retargeted from `feature/tool-drawer-search-grouping` to `develop`.

#1031 merged into its already-merged base branch rather than into `develop`, so the search+grouping commit stranded — see #1033, which re-lands it unchanged. Until #1033 merges this PR shows both commits; afterwards it narrows to just `914d533d` (connection state). Review that commit.

PR #1083, opened 2026-09-13
style(spa): unify page chrome across Agents, Customize and Artifacts

philmerrell:
## Why

These three surfaces had drifted into two different design languages. Most visibly, switching tabs inside a single hub changed the title size and the content width: `/agents` used a `max-w-6xl` shell with a `tracking-tight sm:text-3xl` title, while `/agents/discover`, `/agents/pinned` and all three `/customize` tabs used the `max-w-5xl` + `text-2xl/8` idiom that the admin tables use.

The larger header wasn't a one-off — `/my-skills`, `/schedules` and `/memory-spaces` had already adopted it. Discover, Pinned and Customize were simply the pages that hadn't caught up.

## What changed

**Page header** — seven pages now share `max-w-6xl`, an `mt-6 mb-10` header block, `text-2xl font-bold tracking-tight sm:text-3xl`, and a `mt-1.5 max-w-2xl text-sm/6 text-gray-600` subtitle.

**Pill tabs** — the Artifacts All/Yours/Shared strip was the outlier: full-width, white shell, solid brand-fill active pill. It now uses the same recessed `bg-gray-50` shell and raised white active pill as `AgentsTabsComponent` / `CustomizeTabsComponent`, and `inline-flex` so it hugs its content instead of reading as a segmented control over the whole page. Its grid/list toggle was one radius step larger than the Agents one; they now match.

**Buttons** — there were four different primaries across the three views, including a pre-redesign `rounded-sm` + `focus:ring-3` on Connectors. All are now `rounded-xl bg-primary-accessible px-4 py-2.5 text-sm/6 font-semibold`. `rounded-xl` puts them between the `rounded-2xl` cards and the `rounded-full` chips instead of squared-off against both. The `bg-primary-500` → `primary-accessible` swap is a visual no-op today (both `#0033a0`) but is the only fill guaranteed AA against white text after a rebrand.

**Fields** — Artifacts search and type filter adopt the pill field Discover and Customize already use.

**Grid** — Customize cards go 3-up only at `2xl`. At `lg` the denser toggle cards left ~230px for the title and truncated names like "Student MyBoiseState".

## Bug fixed in passing

Below `sm`, the Artifacts type-select wrapper stretched to full width while the native select stayed shrink-to-fit, stranding the overlaid chevron at the far right, visibly detached from the control. Added the `self-start sm:self-auto` guard the header view toggle already carries — the same failure mode its existing comment describes.

## Docs

`.claude/skills/tailwind-ui/references/app-conventions.md` documented only the admin list/form idiom, so the next session restyling one of these pages would have reverted this. Added a "Top-level user-facing pages" section with the token table, the pill-tab idiom, and the one place `rounded-lg` is still correct (segmented-control children inside a `rounded-xl` shell).

## Testing

- `npx ng test --watch=false` → **2844 passed / 239 files**.
- Browser-verified all three views against real dev data in light and dark, at 1280px and 420px. Light mode checked with both levers (`documentElement.className` empty *and* `prefers-color-scheme`), and the button confirmed by computed style rather than by eye: `border-radius: 12px`, `background: rgb(0, 51, 160)`, `font: 14px/24px`.
- The `2xl` grid change came directly out of that browser pass — the truncation was invisible in isolated-token screenshots.

No behavior change; classes and comments only.

## Follow-up

~47 more primary buttons elsewhere in the app are still on `rounded-lg` / `rounded-sm` / `rounded-md` (admin pages, agent-form, session share modal, fine-tuning, knowledge-base). Being handled separately — that sweep has to settle whether admin keeps its documented `rounded-2xl` or the whole app moves to one radius.

### DIFF
--- a/.claude/skills/tailwind-ui/references/app-conventions.md
+++ b/.claude/skills/tailwind-ui/references/app-conventions.md
@@ -26,10 +26,78 @@ Every token has a dark-mode pair (`dark:*`). Test both modes.
 ```html
 <div class="min-h-dvh">
   <div class="mx-auto max-w-5xl px-4 py-8 sm:px-6 lg:px-8">
-    <!-- list pages: max-w-5xl · form pages: max-w-3xl -->
+    <!-- admin list pages: max-w-5xl · form pages: max-w-3xl -->
   </div>
 </div>
 ```
+
+## Top-level user-facing pages
+
+The pages a non-admin lands on from the sidenav — `/agents` (all three tabs),
+`/customize` (all three tabs), `/artifacts`, `/my-skills`, `/schedules`,
+`/memory-spaces` — use a **larger header and a wider shell** than the admin
+tables above. These are destinations, not records-management screens, and the
+`text-2xl/8` admin title reads as a section label rather than a page.
+
+```html
+<div class="min-h-dvh">
+  <div class="mx-auto max-w-6xl px-4 py-8 sm:px-6 lg:px-8">
+    <app-…-tabs />                     <!-- hub tab strip, if the page is in a hub -->
+
+    <div class="mt-6 mb-10">
+      <h1 class="text-2xl font-bold tracking-tight text-gray-900 sm:text-3xl dark:text-white">
+        Title
+      </h1>
+      <p class="mt-1.5 max-w-2xl text-sm/6 text-gray-600 dark:text-gray-400">
+        One sentence on what the page is for.
+      </p>
+    </div>
+  </div>
+</div>
+```
+
+| Element | Token |
+|---------|-------|
+| Shell | `mx-auto max-w-6xl px-4 py-8 sm:px-6 lg:px-8` |
+| Header block | `mt-6 mb-10` (drop `mt-6` when no tab strip sits above it) |
+| `h1` | `text-2xl font-bold tracking-tight text-gray-900 sm:text-3xl dark:text-white` |
+| Subtitle | `mt-1.5 max-w-2xl text-sm/6 text-gray-600 dark:text-gray-400` |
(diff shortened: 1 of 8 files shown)


philmerrell:
Closing as superseded — not abandoned. Every change here is already carried on `claude/gracious-shtern-9df427`, which imported this work as its first commit (47e93316) and then swept the remaining ~50 off-token buttons across the rest of the app.

Two things changed after this PR was opened:

1. **The radius decision was reversed.** This PR put 6 buttons on `rounded-xl`. The follow-on sweep settled the split the other way — `rounded-2xl` as the single app-wide button radius, matching the ~31 admin pages and dialogs that already used it. That branch retargets these 6 accordingly.
2. **The sweep found more than this PR's inventory did** — 3 buttons where radius and fill weren't adjacent in the class string (login, first-boot, user-detail), plus 6 dialog Cancel buttons and the shared `ConfirmationDialog` pair.

I diffed the two branches file by file: they are identical except the 6 retargeted buttons and one grid breakpoint. That breakpoint is the only thing that would have been lost, so it has been handed to the sweep session explicitly:

```diff
- <ul class="mt-3 grid gap-4 sm:grid-cols-2 lg:grid-cols-3">
+ <ul class="mt-3 grid gap-4 sm:grid-cols-2 2xl:grid-cols-3">
```

in `customize-tools.page.ts` and `customize-skills.page.ts`. At `lg`, with the sidenav open at 1280px, the denser Customize cards leave ~230px for the title and truncate real tool names ("Student MyBoiseS…", "Salesforc…"). That regression only showed up against real dev data — it was invisible in isolated-token screenshots.

One correction for the record: this PR's body cited a 2844-test baseline. That came from an older `develop`; the correct figure is 2824, before #1081 and #1082 landed.

The branch is being left in place until the sweep PR is up and confirmed to carry the grid fix.

PR #1112, opened 2026-09-14
perf(ci): run infra jest across all runner cores

DerrickF:
## What

Bumps the infrastructure jest job from `--maxWorkers=2` to `--maxWorkers=100%`. With the backend suite parallelized in #1111, the infra job became the PR long pole (~6m30s). The old cap left half the runner's cores idle — jest's own default on a 4-vCPU box is 3 workers, so 2 was *below* default.

## Why it's safe
- ts-jest + CDK template synthesis is CPU-bound and each test file is independent → parallelizes cleanly.
- ubuntu-24.04 = 4 vCPU / 16 GB, so 4 workers get ~4 GB each — ample for CDK synth.
- One-line workflow flag, fully reversible.

## Measurement
Baseline infra job: **~6m32s** (`--maxWorkers=2`). Target: this PR's own `Test infrastructure (jest)` run — comparing the two is the proof.

Follows #1111 (backend pytest xdist: ~572s → 318s).


DerrickF:
Closing: measured regression — infra jest went 392s (--maxWorkers=2) to 977s (--maxWorkers=100%), 2.5x slower. ts-jest re-runs type-checking per worker with no shared cache and CDK synth is memory-heavy, so extra workers thrash. Correct fix is a cheaper per-worker transform (transpile-only) plus a single shared tsc gate, not more workers.

PR #1117, opened 2026-09-15
Release/1.22.0

philmerrell:
Release **v1.22.0** — cut from `develop` at `deeddbf4`, carrying 16 PRs since v1.21.0.

## Pre-merge gates

| Gate | Result |
|---|---|
| **GSI update-limit** (§1) | ✅ PASSED — 27 tables compared; no existing table needs more than one GSI operation. `gsi-inventory.json` is unchanged in this range. |
| **Pending backfills** (§1b) | ✅ PASSED — no `backend/scripts/backfill_*.py` added. |
| **`sync-version.sh --check`** | ✅ PASS — all manifests and lockfiles at 1.22.0. |

## What's in it

**Clarifying questions**, end to end — the agent pauses an ambiguous turn, the SPA renders a multiple-choice picker inline, the answer resumes that same tool call, and a refresh rehydrates a pending prompt. The tool worked from PR-2; the model reached for it 4/24 on deliberately ambiguous requests. Rewording the description (44–56%, inside a 17–44% baseline band) and moving its position in the list (25–38%) both stayed in the noise. A system-prompt clause — appended only when the tool is in the turn's **post-filter** effective list — took it to **24/24 ambiguous, 0/18 clear**.

**Admin cost drill-down** — user → conversations → session profile, with 15 diagnosis rules, a per-call context trajectory chart and a copyable diagnostic JSON. **Content-free by construction**: a denylist over the session/cost/upload row families, enforced by a test that walks every admin cost response model (one named exemption) and a moto test that seeds real content and proves none returns. Backed by a new content-free tool census and compaction counter — additive attributes on rows the turn already writes, so no table, no index, no backfill, and nothing reaches the prompt.

**Two silent data bugs**, both invisible / cumulative / delayed:
- Deleting a KB document **mid-upload stranded its byte reservation forever** — every cancelled upload permanently shaved bytes off that assistant's allowance, surfacing months later as "uploads stopped working" with no failure near the deletes that caused it.
- **Born-managed provisioned over an established legacy agent** — legacy KBs share one S3-Vectors index and write no `KB_Record`, so an established agent looked new and its *next* upload flipped retrieval to an empty managed KB, stranding the corpus.

**Generated-document links** work again. The tool result handed the model a ~1,400-char presigned S3 URL that it re-emitted truncated at the `?`. Signed URLs no longer go anywhere they can be copied or persisted — which also drops that tool result from ~1,500 to ~330 characters, in the cacheable prefix, for the life of the session.

## ⚠️ Breaking change — `@`-mention binds the conversation

A mention no longer runs one turn and reverts. Empty thread → the Agent **binds** it; thread with messages → the message opens a **new conversation** with that Agent. This reverses design decision D11 on new evidence: of **247 prod mentions, 247 started the conversation** (dev: 60/61). It retires a failure where the Agent's tools vanished silently after turn one (`Unknown tool: create_rubric`, with the model confidently blaming a setting that was already correct), plus the ~$0.12-per-mention prefix re-write and the history fork. No migration needed.

## 🚀 Deployment

**Skip `platform.yml`** — the only `infrastructure/` diff is `jest.config.js` and the version bump. Deploy `backend.yml` → `frontend-deploy.yml`.

**One required operator step, per environment: enable the Clarifying Questions tool.** `seed_bootstrap_data.py` skips any tool row that already exists, so the seed's `enabledByDefault: True` reaches a fresh bootstrap only. Where the row exists, an admin must flip it on; where it does not, run the seed. Roles without a `*` tool grant need `ask_user_question` added to the **role record** (the `allowedAppRoles` projection grants nothing). Full instructions in the release notes.

Two new env vars, both **default on when unset** (`ASK_USER_QUESTION_ENABLED`, `COST_DIAGNOSTICS_ENABLED`), so neither needs plumbing to ship.

## Also in this release

- KB storage usage bar with byte cap on the agent's KB card (managed only — legacy is uncapped)
- Log-injection sinks sanitized across 9 modules; nightly ref allowlist pinned by a test
- CI: backend pytest parallelized with xdist (`-n auto`); infra jest transpile-only with a separate `tsc --noEmit` preserving type safety
- 5,500+ lines of new tests — 17 backend modules, 9 frontend specs

Full detail: [RELEASE_NOTES.md](RELEASE_NOTES.md) · [CHANGELOG.md](CHANGELOG.md)

---

**Merge with a squash.** After merge, the backmerge `main` → `develop` is required (merge commit, not squash).


philmerrell:
## CodeQL triage — 1 alert left, and it is pre-existing

The `CodeQL` check is red. It is the only red check; the other 14 pass. Two alerts were reported, and they needed opposite responses — posting the triage so a reviewer doesn't re-derive it.

### ✅ Fixed — `py/log-injection` (medium, alert 864) was a real regression

`backend/src/apis/app_api/documents/routes.py:129` interpolated `assistant_id` — a user-controlled path parameter — straight into a `logger.warning` f-string, so CR/LF in it could forge log lines.

It arrived with the KB storage usage bar (8fb4f8f4, #1108), on a branch parallel to the #1098 sweep in this same release that wrapped every *other* such sink. So the release was about to ship a regression against a rule it enforces elsewhere. Fixed in 3e300729 to the sweep's own convention — `%s` lazy formatting with `scrub_log()` on each user-influenced value. The check summary went from `2 new alerts including 1 high` to `1 new alert including 1 high`.

I also scanned every logger call this release **adds** for the same shape; the rest carry exceptions, ints, or already-scrubbed values. The pre-existing f-string logger calls elsewhere in the tree are deliberately untouched — that is a sweep, not release-branch work.

### ⚠️ Not a regression — `actions/cache-poisoning/poisonable-step` (high, alert 863)

`.github/workflows/tests.yml:157`. **This alert does not represent new risk, and it will not clear by editing this branch.**

`main` already has **13 open alerts of this exact rule family**, verified just now:

| File | Count | Alerts | Lines |
|---|---|---|---|
| `.github/workflows/nightly.yml` | 8 | 800–806, 810 | 267, 294, 298, 302, 336, 377, 433, 600 |
| `.github/workflows/tests.yml` | **5** | 807, 809, 847, 848, 849 | 83, 86, 129, 132, 151 |

```bash
gh api "repos/Boise-State-Development/agentcore-public-stack/code-scanning/alerts?ref=refs/heads/main&state=open&per_page=100" \
  --jq '[.[] | select(.rule.id|test("cache-poisoning"))] | group_by(.most_recent_instance.location.path) | .[] | {file: .[0].most_recent_instance.location.path, count: length}'
```

The rule flags every *step* that runs where a poisoned cache could have been restored. The `Test infrastructure (jest)` job already carries the pattern on `main` — `ref: ${{ inputs.ref }}` at line 144, `cache: 'npm'` at 149 — and its `Install` step at line 151 is **already flagged as alert 809**.

This release adds one step to that job, the `Type-check (tsc --noEmit)` step from #1113 that preserves type safety now that ts-jest transpiles only. That new step becomes alert 863. **Same job, same pre-existing pattern, one more step — the step count changed, the risk did not.**

The actual mitigation is that `nightly.yml`, the only privileged caller passing a `ref`, resolves track tokens through a shell `case` assigning literal `main` / `develop` and hard-`exit 1`s on anything else. CodeQL cannot see through a shell `case`, which is stated outright in the docstring of `backend/tests/supply_chain/test_nightly_ref_allowlist.py` — the guard added in **this release** (#1098) that pins the allowlist so widening it can't silently make the standing finding real.

### Recommendation

Merge. The one real finding is fixed; the remaining high is an accepted, tested, pre-existing pattern that this release inherits rather than introduces. `main` is not branch-protected, so the red check does not block — but it is a knowing call, hence this comment.

Worth noting for the reviewer: `codeql.yml` only runs on `main`, so a release PR is always the first time a release's code meets CodeQL. Triage each alert against `main`'s own baseline rather than treating the batch as one verdict — this PR is exactly the case where a real finding and a stale one arrived together, and the stale one is what tempts you to wave both through.

PR #1152, opened 2026-09-18
feat(feedback): fleet-level attribution — down-thumb rate by config arm (response-feedback PR-3)

philmerrell:
Targets `develop`. Originally stacked on #1148 for the `FEEDBACK#down` / `FEEDBACK#up` queue keys it reads; #1146 and #1148 have both merged, so this is now re-targeted and develop is merged in.

Completes `docs/specs/response-feedback.md` §11 PR-3, and answers the compaction spec's §7.2 request for an outcome signal joined to compaction.

## Why

Everything built so far is per conversation. An admin can inspect feedback only for a session they already found some other way, almost always by cost, so "where are users unhappy?" has had no answer in the console. §7 calls this join the prize: down-thumb rate against the configuration choices we actually made.

## What

`GET /admin/feedback/fleet?days=` reads a trailing window off the two feedback partitions on `UserTimestampIndex` (a range read, no new index), joins each thumb to its `C#` cost row, and splits it four ways:

| Dimension | Question it answers |
|---|---|
| **model** | Is a cheaper model costing answer quality on this workload? |
| **distance from a compaction cut** | Does quality drop after history is summarised away? Bucketed in model calls: never / same call / 1-3 / 4+ |
| **agent switch** | An `@`-mention forks the history. Does it also cost quality? |
| **document turn class** | Full document inline, digest only, or pages retrieved back |

`/admin/feedback` renders it under the `admin.costs` scope, with a 7/30/90 day window, a bar per arm, the reason-code distribution, and a coverage footer.

## How §9 is enforced rather than promised

The spec is emphatic that this surface must not become a quality KPI, so the rules are structural:

- **Every arm carries its own `n`.** A rate never travels without one.
- **An arm under the floor returns `downRate: null`**, not a number the UI could round. There is nothing to quote. The page renders "n < 20" in grey.
- **No fleet-wide rate field exists at any level.** `totals` carries counts. The page's headline comparison is the gap between the two furthest-apart *reportable* arms, which is the comparison §9 asks for, and it is absent entirely when fewer than two arms qualify.
- **Coverage is reported, not implied**: thumbs seen, thumbs joined, join rate, conversations, and any truncation.
- **Caps are honest.** 2,000 thumbs and 300 joined sessions per query; exceeding either sets `coverage.truncated` / `sessionsOmitted` and the page says the arms rest on a subset.

**Open question 3 is answered.** The coverage floor is **20**, picked because below it one extra thumb swings the rate by more than five points, which is coarser than any decision would want. It is a named constant with an env override (`FEEDBACK_ARM_MINIMUM_N`), not a literal.

## Two deviations from §11 PR-3 as written

- **A page of its own, not a panel beside the cost panels.** The cost drill-down is per conversation; this question is fleet-shaped. A panel there would have ignored the session whose page it sat on.
- **No skills arm.** `enabledTools` lives on the `S#` session row, not the `C#` call row, so a per-call skills arm needs a write-time denormalisation or a second join. Not worth doing before the signal is proven. Recorded in §13.

## What this does not change

- No new index, no CDK, no flag. The route rides the existing `admin.costs` scope.
- Nothing on the chat path. This is read-only over rows already being written.
- **Implicit signals are invisible here by construction**: `record_implicit_signal` writes no `GSI1*` keys, so the window query cannot see them (verified against merged develop, not just assumed). The `signal` filter in the aggregator is belt-and-braces. Surfacing them in the fleet view is a follow-up.
- `_turn_class` in the costs service is renamed to a public `turn_class` so the fleet module shares it rather than duplicating the precedence rule. No behaviour change.

## Tests

- **Arm math** (12 tests): below-floor withholding, env-tunable floor that can never be zero, compaction distance bucketing including the "never" control arm and an unplaceable message, last-row-wins when a turn's tool round trips share a `messageId`, full report with unjoined thumbs and an implicit row excluded, compaction arm keeping distance order while others lead with the biggest sample, and an explicit assertion that **no fleet-wide rate field exists**.
- **Route**: windowing and the reported window matching what storage was asked for, truncation and omitted sessions, a session whose rows fail to read degrading to `unjoined` rather than failing, bounds on `days`, 500 mapping, and a content-policy walk over the response body.
- **Storage** (moto): both polarities in one window, time bounds excluding rows either side, no user id and no content in the projection.
- **SPA**: util spec (6) and page spec (7), including that a 33% fleet rate derivable from the totals is never rendered, that switching the window reloads, the empty state, and the read-cap warning. The `admin-scope-wiring` spec passes, which is what validates the new route and nav entry share a scope.
- Full backend suite before the develop merge: **9054 passed, 3 skipped**. After merging develop (which brought #1146 and #1148): the affected backend set re-ran at **475 passed**, the SPA at **79 passed** across 6 files, and `ng build` (AOT) is clean; the full suite re-run is reported in a comment below.

### Merge conflict resolution

Only `docs/specs/response-feedback.md` conflicted, on the status line, where three PRs each claimed a different set of completed phases; the resolution names all of them. Code auto-merged, and the two places that could have merged cleanly but wrongly were checked by hand: no stale `_turn_class` call sites survive the rename to public `turn_class`, and `record_implicit_signal` still writes no `GSI1*` keys, which is what keeps implicit rows out of this window query.
- **Not exercised in a browser.** Dev has no accumulated feedback yet, so the page would render its empty state; the component spec covers the populated shapes instead.


philmerrell:
Full backend suite on the merged tree (develop, including #1146 and #1148): **9192 passed, 3 skipped**.

